PR #25, opened 2017-02-02
i965: remove unnecessary code from i965_Terminate when compiling without HAVE_HYBRID_CODEC

tmm1:
Minor fix for what looks like a simple oversight.

### DIFF
--- a/src/i965_drv_video.c
+++ b/src/i965_drv_video.c
@@ -6720,6 +6720,7 @@ i965_Terminate(VADriverContextP ctx)
     int i;
 
     if (i965) {
+#if HAVE_HYBRID_CODEC
         if (i965->wrapper_pdrvctx) {
             VADriverContextP pdrvctx;
             pdrvctx = i965->wrapper_pdrvctx;
@@ -6732,6 +6733,7 @@ i965_Terminate(VADriverContextP ctx)
             free(pdrvctx);
             i965->wrapper_pdrvctx = NULL;
         }
+#endif
 
         for (i = ARRAY_ELEMS(i965_sub_ops); i > 0; i--)
             if (i965_sub_ops[i - 1].display_type == 0 ||


src/i965_drv_video.c
```diff
@@ -6720,6 +6720,7 @@ i965_Terminate(VADriverContextP ctx)
     int i;
 
     if (i965) {
+#if HAVE_HYBRID_CODEC
```
uartie: Is this really necessary since we check for i965->wrapper_pdrvctx.  I know, it's a compile-time check vs. runtime check... but we're in i965_Terminate and don't think this change will give us much gain.  If we do this compile-time check, why not change it everywhere else we check for i965->wrapper_pdrvctx?

uartie: What are you trying to fix?

tmm1: No, it is not strictly necessary. The reason I ran into this is because I'm compiling the intel driver as a static library and the invocation of `dlclose()` in this code segment was causing problems.

(I have a patched version of libva which allows it to be built and used statically with the intel driver embedded.)

uartie: That would be good information to put in the commit message and file an issue for.

uartie: Context and reasoning to support your change will make other developers think twice before attempting to revert it in the future ;-)

tmm1: Fair enough, I have updated the commit message accordingly.

seanvk:
@tmm1 Please file a bug, and add fixes # to the commit log, otherwise looks fine.

Sean

tmm1:
The merge commit will reference #25 already, do I really need to duplicate into a new issue?

seanvk:
https://github.com/01org/intel-vaapi-driver/blob/master/CONTRIBUTING.md

Please file an issue, update the commit header with fixes #  

We do this for QA and tracking.

Thanks,

Sean

tmm1:
Done.

seanvk:
lgtm, thanks.

PR #32, opened 2017-02-07
vdenc fixes

yakuizhao:


src/gen9_vdenc.c
```diff
@@ -3456,14 +3475,16 @@ gen9_vdenc_mfx_vdenc_avc_slices(VADriverContextP ctx,
                     insert_mi_flush = 0;
                 }
 
+                pipeline_flush_params.mfx_pipeline_done = 1;
```
xhaihao: Please fix the assignment to pipeline_flush_params.mfx_pipeline_done instead of a forced assignment

src/gen9_vdenc.c
```diff
@@ -3670,13 +3670,13 @@ gen9_vdenc_avc_encode_picture(VADriverContextP ctx,
...
-
         vdenc_context->brc_initted = 1;
         vdenc_context->brc_need_reset = 0;
     }
 
+    intel_batchbuffer_end_atomic(batch);
+    intel_batchbuffer_flush(batch);
+
     return VA_STATUS_SUCCESS;
 }
```
xhaihao: Could you explain the reason of conflicting in the commit log?

src/gen9_vdenc.c
```diff
@@ -3120,6 +3133,9 @@ gen9_vdenc_mfx_avc_inset_headers(VADriverContextP ctx,
 
     if (vdenc_context->is_frame_level_vdenc)
         insert_one_zero_byte = 0;
+    else {
+        insert_one_zero_byte = 1;
+    }
```
yakuizhao: The flag of insert_one_zero_byte will vary on the platform.
So I would like to keep the current flowchart. When it is for is_frame_level_vdenc==0, it is configured again.

For the flag check of gen9_vdenc_mfx_avc_insert_slice_packed_data(), currently we only prepare one slice_header for every slice.  If multi data buffers are needed for one slice, it is not clear whether it is really needed.
I will double check it.

src/gen9_vdenc.c
yakuizhao: Sure. This will be fixed in next version.

src/gen9_vdenc.c
yakuizhao: The flag of insert_one_zero_byte will vary on the platform.
So I would like to keep the current flowchart. When it is for is_frame_level_vdenc==0, it is configured again.

For the flag check of gen9_vdenc_mfx_avc_insert_slice_packed_data(), currently we only prepare one slice_header for every slice.  If multi data buffers are needed for one slice, it is not clear whether it is really needed.
I will double check it.

src/gen9_vdenc.c
```diff
@@ -2961,6 +2962,13 @@ gen9_vdenc_mfx_avc_insert_slice_packed_data(VADriverContextP ctx,
 
         /* No slice header data is passed. And the driver needs to generate it */
         /* For the Normal H264 */
+
+        if (slice_index &&
+            IS_KBL(i965->intel.device_info)) {
+
+            slice_params = (VAEncSliceParameterBufferH264 *)encode_state->slice_params_ext[0]->buffer;
```
yakuizhao: Yes.  But it is only used for creating the slice_header buffer.

src/gen9_vdenc.c
yakuizhao: The main purpose is to avoid the issue that the status register in previous is not stored while it is used in the subsequent pass.  It seems that the current code already stores the register into the internal buffer. In such case the commit log is incorrect.

Of course the patch is to avoid submitting batch buffer multi times. When it works in multi-pass scenarios, maybe the first pass already meets with the requirement. In such case subsequent pass is not necessary.  But the current code still prepares it and submits it. It is not meaningful.

seanvk: okay, had we not considered the multi-pass scenario previously?

src/gen9_vdenc.c
```diff
@@ -2810,11 +2810,11 @@ gen95_vdenc_vdenc_walker_state(VADriverContextP ctx,
     int luma_log2_weight_denom, weighted_pred_idc;
 
     slice_hor_pos = slice_param->macroblock_address % vdenc_context->frame_width_in_mbs;
-    slice_ver_pos = slice_param->macroblock_address / vdenc_context->frame_height_in_mbs;
+    slice_ver_pos = slice_param->macroblock_address / vdenc_context->frame_width_in_mbs;
```
seanvk: What do you mean by wrong?  We shouldn't be using y-ccord for slice vertical position?  Why?

src/gen9_vdenc.c
seanvk: The header for slice index zero makes sense.  I'm not so sure about whether multi data buffers are needed.

src/gen9_vdenc.c
seanvk: agree.

src/gen9_vdenc.c
```diff
@@ -3456,14 +3475,16 @@ gen9_vdenc_mfx_vdenc_avc_slices(VADriverContextP ctx,
...
                 pipeline_flush_params.vdenc_pipeline_command_flush = 1;
                 pipeline_flush_params.vd_command_message_parser_done = 1;
                 gen9_vdenc_vd_pipeline_flush(ctx, encoder_context, &pipeline_flush_params);
 
                 if (insert_mi_flush) {
                     memset(&mi_flush_dw_params, 0, sizeof(mi_flush_dw_params));
-                    mi_flush_dw_params.video_pipeline_cache_invalidate = 1;
+                    mi_flush_dw_params.video_pipeline_cache_invalidate = 0;
+                    mi_flush_dw_params.bo = vdenc_context->vdenc_internal_res.bo;
                     gen8_gpe_mi_flush_dw(ctx, batch, &mi_flush_dw_params);
```
seanvk: Is this new to KBL?  So we will need it for KBL+.

src/i965_device_info.c
```diff
@@ -440,6 +441,7 @@ static struct hw_codec_info kbl_hw_codec_info = {
     .render_init = gen9_render_init,
     .post_processing_context_init = gen9_post_processing_context_init,
     .max_resolution = gen9_max_resolution,
+    .preinit_hw_codec = gen9_hw_codec_preinit,
```
seanvk: Are we already doing this for SKL?

src/gen9_vdenc.c
xhaihao: How about if slice 0 and other slices use a different slice qp delta?

src/gen9_vdenc.c
yakuizhao: The AVC_SLICE_STATE GPU commands need the two kind of coord for the slice. 
       >slice_x/slice_y for the current slice
       >slice_x/slice_y for the next slice

So we need to calculate them based on macroblock_address. 
The current calculation is wrong especially for multi-slices. Of course the issue is hidden for the single slice that the macroblock_address is zero.

yakuizhao: The AVC_SLICE_STATE GPU commands need the two kind of coord for the slice. 
       >slice_x/slice_y for the current slice
       >slice_x/slice_y for the next slice

So we need to calculate them based on macroblock_address. 
The current calculation is wrong especially for multi-slices. Of course the issue is hidden for the single slice that the macroblock_address is zero.

src/gen9_vdenc.c
yakuizhao: Yes. For the SKL/BXT, only one single slice is supported.
So it is not necessary for SKL.
But for KBL, this is needed for multi-slices.

src/i965_device_info.c
yakuizhao: Yes. It will check the Huc attribute in the mentioned callback functions so that the BRC rate_control is returned.
Otherwise NO other BRC rate_control except CQP is returned

src/gen9_vdenc.c
yakuizhao: Multi-pass scenario is also considered and there is no problem.
This patch is only for the optimization purpose.  Previously, maybe multi drm_ioctls are needed.
After it is combinted, only one drm_ioctl is needed.

xhaihao: Yakui's fix is trying to use one ioctl call but with a big batchbuffer to replace two ioctl calls with small batchbuffers. A big batchbuffer consumes more GPU resource and might delay the execution of other batchbuffers from other applications . On the other hand, I think the increase in performance is trivial because the commands for the 2nd pass is also submitted to kernel for the case of big batchbuffer.  Note we can not decide the execution of the 2nd pass in CPU side,  otherwise CPU has to check the result of the first batchbuffer , which implies a sync between CPU and GPU and will impact the performance a lot.

src/gen9_vdenc.c
xhaihao: Yes, the calculation of the start position of the slice in the driver is wrong.

src/i965_device_info.c
xhaihao: Actually there is a hidden bug when creating a VA config with VAConfigAttribRateControl. The current driver can do CBR and VBR even if the corresponding attribute is not returned to user via  VAConfigAttribRateControl.

yakuizhao: Agree. In fact the issue not only exists for VDENC, but also exits for the other codecs.(H264/HEVC/VP8/VP9).
The problem is related with that vaCreateConfig/vaCreateContext doesn't check whether the input rate_control is supported. 
The driver needs to be enhanced to improve the robustness.

seanvk: @xhaihao yes, and let's file a bug for that on github please.

xhaihao: @seanvk Yakui filed https://github.com/01org/intel-vaapi-driver/issues/35 to track this issue.

src/gen9_vdenc.c
xhaihao: Why a write is required here? flush can be done without write.

xhaihao:
@yakuizhao please respond to Sean's requested changes. You could dismiss the review after providing the new version with the changes

> Please follow-up with questions.
Please create a Github issue.
Please add "Fixes #n" to the commit headers.
Please provide description of issues for QA to verify in the issue.
If necessary, create separate issue for each commit whether they address specific problems for QA to check and in so doing add the Fixes #n to the commit header.

src/gen9_vdenc.c
seanvk: @yakuizhao Could you follow-up here?

yakuizhao: The write is not required.
It will be removed in next version.

PR #33, opened 2017-02-07
AVC Encoder Architecture rework

QuPengfei:
Encoder Architecture Changes (Primarily AVC)

Encoder architecture restructuring for H.264 (with some impact to HEVC now) on HSW+
* Improvements to the shaders
* Improvements to the B frame efficiency
* Improvements to the low bit rate mode
* Improved features in two stage VME/PAK pipeline

v1:
Reduce the patch number and re org for VME and MFX related patches.
Patch re org for VME pipeline
Patch re org for MFX pipeline
keep assert for internal logic and replace assert for input validation function.
Remove unnecessary comments and enum value.
Use the 64bit version OUT_BCS_RELOC64.
Move kernel binary into header file.
use misc parameter from encoder_context structure.

v2:
add APL support
For each patch, new file is added in the Makefile.am
Fix zero divide issue under cqp
Correct the license header for the new file.
Fix the Batch buffer logic when conditional end enabled.
Remove the execution bit for the files.
misc changes.

v3:
rebase to lastest master branch and use i965->intel->media_mocs to configure the memory MOCS.
remove unused enum value 

v4:
use i965->intel->media_mocs to configure the memory MOCS in PAK pipeline

Reviewed-by: Sean V Kelley<seanvk@posteo.de>
Reviewed-by: yakuizhao <yakui.zhao@intel.com>


src/i965_encoder_common.h
```diff
@@ -0,0 +1,517 @@
...
+    INTEL_PRESET_OPT_SPEED       = 3,
+    INTEL_PRESET_RT_SPEED        = 4,
+    INTEL_PRESET_HI_SPEED        = 6,
+    INTEL_PRESET_BEST_SPEED      = 7,
+    INTEL_PRESET_LOW_LATENCY     = 0x10,
+    INTEL_PRESET_MULTIPASS       = 0x20
+};
+/*
+   the definition for encoder status
+*/
```
yakuizhao: As this is mentioned in previous review cycle, the following two structures vary on the different platforms/codec.
It doesn't make sense that it is added as the generic structure.
Otherwise we will have to be changed frequently.

src/i965_encoder_common.h
```diff
@@ -0,0 +1,517 @@
...
+    INTEL_ENCODE_BRCUPDATE_IS_FIELD                  = 0x01,
+    INTEL_ENCODE_BRCUPDATE_IS_MBAFF                  = (0x01 << 1),
+    INTEL_ENCODE_BRCUPDATE_IS_BOTTOM_FIELD           = (0x01 << 2),
+    INTEL_ENCODE_BRCUPDATE_IS_ACTUALQP               = (0x01 << 6),
+    INTEL_ENCODE_BRCUPDATE_IS_REFERENCE              = (0x01 << 7)
+} INTEL_ENCODE_BRCUPDATE_FLAG;
+
+/*
+kernel operation related defines
+*/
```
yakuizhao: The following two definitions still include a lot of definitions that is not used.
For example: MBENC_WIDI, VDENC_ME, RESET_VLINE_STRIDE and so on.,
Please double check it and remove the unused/unmeaningful definition.

yakuizhao: The following two definitions still include a lot of definitions that is not used.
For example: MBENC_WIDI, VDENC_ME, RESET_VLINE_STRIDE and so on.,
Please double check it and remove the unused definition.

src/i965_avc_encoder_common.h
```diff
@@ -0,0 +1,305 @@
...
+};
+
+typedef enum
+{
+    INTEL_AVC_BASE_PROFILE               = 66,
+    INTEL_AVC_MAIN_PROFILE               = 77,
+    INTEL_AVC_EXTENDED_PROFILE           = 88,
+    INTEL_AVC_HIGH_PROFILE               = 100,
+    INTEL_AVC_HIGH10_PROFILE             = 110,
+    INTEL_AVC_HIGH422_PROFILE            = 122,
```
yakuizhao: Please remove the  misleading definition(for example: HIGH422, HIGH444 and so on).
Otherwise maybe someone think that it will be supported in the driver.

src/gen9_avc_encoder.c
```diff
@@ -0,0 +1,6027 @@
...
+    pstate->dw4.insert_test_flag = 0;
+    pstate->dw4.load_slice_pointer_flag = 0;
+    pstate->dw4.macroblock_stat_enable = 0;        /* disable in the first pass */
+    pstate->dw4.minimum_frame_size = 0;
+    pstate->dw5.intra_mb_max_bit_flag = 1;
+    pstate->dw5.inter_mb_max_bit_flag = 1;
+    pstate->dw5.frame_size_over_flag = 1;
+    pstate->dw5.frame_size_under_flag = 1;
+    pstate->dw5.intra_mb_ipcm_flag = 1;
+    pstate->dw5.mb_rate_ctrl_flag = 0;             /* Always 0 in VDEnc mode */
```
yakuizhao: Still keep the VDENC comment?

src/gen9_avc_encoder.c
```diff
@@ -0,0 +1,6027 @@
...
+    struct encoder_vme_mfc_context * pak_context = (struct encoder_vme_mfc_context *)encoder_context->vme_context;
+    struct generic_enc_codec_state * generic_state = (struct generic_enc_codec_state * )pak_context->generic_enc_state;
+    char *pdata;
+    int i;
+    unsigned int * data;
+    struct gen9_mfx_avc_img_state cmd;
+
+    pdata = i965_map_gpe_resource(gpe_resource);
+
+    gen9_avc_init
```
yakuizhao: As it is pointed in previous review cycle, the incorrect ref_count is used for avc_priv_surface->dmv_top/dmv_bottom.
Please double check it again.

yakuizhao: The same issue of reference count.

src/gen9_avc_encoder.c
```diff
@@ -0,0 +1,6027 @@
...
+            generic_state->brc_need_reset = 1;
+        }
+    }
+
+    /*  frame rate */
+    if (generic_state->internal_rate_mode != VA_RC_CQP)
+    {
+        generic_state->frames_per_100s = encoder_context->brc.framerate[0].num/encoder_context->brc.framerate[0].den * 100;
+        generic_state->frame_rate = encoder_context->brc.framerate[0].num/encoder_context->brc.framerate[0].den ;
+    }
```
yakuizhao: As it is pointed in previous review cycle, the above code has the precision loss.
Had better be changed to:
    num * 100 / den.

src/gen9_avc_encoder.c
```diff
@@ -4493,14 +4500,14 @@ gen9_avc_set_curbe_sfd(VADriverContextP ctx,
     memset(cmd,0,sizeof(gen9_avc_sfd_curbe_data));
 
     cmd->dw0.enable_intra_cost_scaling_for_static_frame = 1 ;
```
yakuizhao: The VDENC comment had better be removed in previous patch.
Otherwise it is changed back and forth.

QuPengfei: it will be removed

src/gen9_avc_encoder.c
QuPengfei: it will be fixed

src/gen9_avc_encoder.c
QuPengfei: it will be fixed

QuPengfei: it will be fixed

src/gen9_avc_encoder.c
QuPengfei: it will be removed

src/i965_avc_encoder_common.h
QuPengfei: HIGH422, HIGH444 will be removed.

src/i965_encoder_common.h
QuPengfei: WIDI is in the next plan. 
Other definitions will be removed.

QuPengfei: WIDI is in the next plan. 
Other definitions will be removed.

src/i965_encoder_common.h
QuPengfei: we have a plan to refine these two structures for all VP8/VP9/AVC/HEVC. it will common for them

src/i965_encoder.c
```diff
@@ -800,12 +803,32 @@ intel_encoder_check_misc_parameter(VADriverContextP ctx,
         VAEncMiscParameterBufferQualityLevel* param_quality_level = (VAEncMiscParameterBufferQualityLevel*)pMiscParam->data;
         encoder_context->quality_level = param_quality_level->quality_level;
 
-        if (encoder_context->quality_level == 0)
-            encoder_context->quality_level = ENCODER_DEFAULT_QUALITY;
-        else if (encoder_context->quality_level > encoder_context->quality_range) {
-            ret = VA_STATUS_ERROR_INVALID_PARAMETER;
-            goto out;
+        switch (profile) {
+        case VAProfileH264ConstrainedBaseline:
```
pengche1: it is common place for quality level and not suitable to put h.264 special codes here.

pengche1: It is the common place for quality level and not suitable to put h.264 special codes here

src/i965_gpe_utils.c
```diff
@@ -1406,6 +1406,27 @@ i965_object_surface_to_2d_gpe_resource(struct i965_gpe_resource *res,
 }
 
 void
+i965_object_surface_to_2d_gpe_resource_with_align(struct i965_gpe_resource *res,
+                                                  struct object_surface *obj_surface)
+{
+    unsigned int swizzle;
```
pengche1: For those special surface, it will be better directly change its orig_width and orig_height to 16 aligned value. So this aligned function is not required and  i965_object_surface_to_2d_gpe_resource() is enough for your requirement

QuPengfei: i965_object_surface_to_2d_gpe_resource() also will be used by other function, such as VPP. it will cause the issue to change the value in the function i965_object_surface_to_2d_gpe_resource(). that is the reason why add the version with aligned

pengche1: keep i965_object_surface_to_2d_gpe_resource() and don't modify it. But update obj_surface->orig_width and  obj_surface->orig_width to new aligned value

QuPengfei: Directly change the surface value(orig_width ,orig_height) looks not good. Here GPE resource is the birdge to attend the final operation. it is better to do alignment with gpe resource. The issue maybe be there when the post process on this surface.
BTW, the best align operation should be done when obj surface is allocated . But this is implemented in the middle ware. For driver, it only work around.

QuPengfei: Directly change the surface value(orig_width ,orig_height) looks not good. Here GPE resource is the birdge to attend the final operation. it is better to do alignment with gpe resource. The issue maybe be there when the post process on this surface.
BTW, the best align operation should be done when obj surface is allocated . But this is implemented in the middle ware. For driver, it only work around.

QuPengfei: Directly change the surface value(orig_width ,orig_height) looks not good. Here GPE resource is the birdge to attend the final operation. it is better to do alignment with gpe resource. The issue maybe be there when the post process on this surface.
BTW, the best align operation should be done when obj surface is allocated . But this is implemented in the middle ware. For driver, it only work around.

src/i965_encoder.c
QuPengfei: move quality level check in the brc_prepare function

src/i965_gpe_utils.c
pengche1: It is resonable and ok for me

xhaihao: i965_object_surface_to_2d_gpe_resource() and i965_object_surface_to_2d_gpe_resource_with_align are almost the same. I suggest take the alignment as an input parameter to i965_object_surface_to_2d_gpe_resource_with_align(), define  i965_object_surface_to_2d_gpe_resource as below
void i965_object_surface_to_2d_gpe_resource(struct i965_gpe_context *gpe_context,
                             struct i965_gpe_surface *gpe_surface,
                             int index)
{
      i965_object_surface_to_2d_gpe_resource_with_align(gpe_context, gpe_surface, 1);
}

We will only modify one function if there is any error in i965_object_surface_to_2d_gpe_resource(),i965_object_surface_to_2d_gpe_resource_with_align(). In addition, we can set other alignments if needed.

src/i965_encoder.c
```diff
@@ -1485,5 +1494,17 @@ gen8_enc_hw_context_init(VADriverContextP ctx, struct object_config *obj_config)
 struct hw_context *
 gen9_enc_hw_context_init(VADriverContextP ctx, struct object_config *obj_config)
 {
-    return intel_enc_hw_context_init(ctx, obj_config, gen9_vme_context_init, gen9_mfc_context_init);
+    struct i965_driver_data *i965 = i965_driver_data(ctx);
+    switch (obj_config->profile){
+    case VAProfileH264ConstrainedBaseline:
+    case VAProfileH264Main:
+    case VAProfileH264High:
```
xhaihao: Miss VAProfileH264MultiviewHigh and VAProfileH264StereoHigh, and you can use encoder_context->codec as conditional.

src/i965_encoder.c
```diff
@@ -1485,5 +1494,17 @@ gen8_enc_hw_context_init(VADriverContextP ctx, struct object_config *obj_config)
...
+        if (IS_SKL(i965->intel.device_info)||
+            IS_BXT(i965->intel.device_info))
+            return intel_enc_hw_context_init(ctx, obj_config, gen9_avc_vme_context_init, gen9_avc_pak_context_init);
+        else
+            return intel_enc_hw_context_init(ctx, obj_config, gen9_vme_context_init, gen9_mfc_context_init);
+    default:
+        return intel_enc_hw_context_init(ctx, obj_config, gen9_vme_context_init, gen9_mfc_context_init);
+    }
```
xhaihao: It would be better to avoid too many platform specific code (IS_xxx() etc) in a generic file, you could move these code to a gen9 related file.

src/i965_drv_video.h
```diff
@@ -69,7 +69,9 @@
 #define DEFAULT_SATURATION      50
 
 #define ENCODER_QUALITY_RANGE     2
+#define ENCODER_QUALITY_RANGE_AVC     8
```
xhaihao: Does the new encoder really support quality level from 1 to 8? I think HIGH(1)/Medium(2, default)/Low(3) is enough.

src/i965_drv_video.c
```diff
@@ -1003,8 +1004,11 @@ i965_GetConfigAttributes(VADriverContextP ctx,
                 attrib_list[i].value = 1;
                 if (profile == VAProfileH264ConstrainedBaseline ||
                     profile == VAProfileH264Main ||
-                    profile == VAProfileH264High )
-                    attrib_list[i].value = ENCODER_QUALITY_RANGE;
+                    profile == VAProfileH264High ){
+                        attrib_list[i].value = ENCODER_QUALITY_RANGE;
+                        if(IS_GEN9(i965->intel.device_info))
```
QuPengfei: that is correct. it will be fixed

src/i965_encoder.c
```diff
@@ -1394,7 +1398,12 @@ intel_enc_hw_context_init(VADriverContextP ctx,
         encoder_context->codec = CODEC_H264;
 
         if (obj_config->entrypoint == VAEntrypointEncSliceLP)
-            encoder_context->quality_range = ENCODER_LP_QUALITY_RANGE;
+            encoder_context->quality_range = ENCODER_QUALITY_RANGE_AVC;
```
QuPengfei: No changes for vdenc. i will revert this change for VDENC.

src/gen9_avc_encoder.c
```diff
@@ -5748,6 +5748,1596 @@ gen9_avc_kernel_init(VADriverContextP ctx,
...
+            avc_state->top_field_poc[2*i] = avc_priv_surface->top_field_order_cnt;
+            avc_state->top_field_poc[2*i+1] = avc_priv_surface->top_field_order_cnt+1;
+        }else
+        {
+            break;
+        }
+    }
+
+    if (avc_ctx->pres_slice_batch_buffer_2nd_level)
+        intel_batchbuffer_free(avc_
```
xhaihao: If not, please remove the code below, otherwise please update the comment

src/i965_drv_video.h
QuPengfei: AVC kernel support quality level 1-7 by now. The test result show 1-3 is closer and 4-6 is closer.

MediaSDK support 1-7.

src/gen9_avc_encoder.c
```diff
@@ -5748,6 +5748,1596 @@ gen9_avc_kernel_init(VADriverContextP ctx,
...
+    memset(&mi_store_reg_mem_param, 0, sizeof(mi_store_reg_mem_param));
+    mi_store_reg_mem_param.bo = status_buffer->bo;
+    mi_store_reg_mem_param.offset = status_buffer->mfc_qp_status_count_offset;
+    mi_store_reg_mem_param.mmio_offset = status_buffer->mfc_qp_status_count_reg_offset;
+    gen8_gpe_mi_store_register_mem(ctx, batch, &mi_store_reg_mem_param);
+
+    mi_store_reg_mem_param.bo = status_buffer->bo;
+    mi_store_reg_mem_param.offset = status_buffer->image_status_mask_offset;
+    mi_store_reg_mem_param.mmio_offset = status_buffer->image_status_mask_reg_offset;
+    gen8_gpe_mi_store_register_mem(ctx, batch, &mi_store_reg_mem_param);
```
xhaihao: Only the first info in status_buffer is used later, I suggest to read status_buffer->bs_byte_count_frame_offset only and delete unused fields.

src/i965_encoder.c
QuPengfei: move it into the function gen9_mfc_context_init and gen9_vme_context_init

src/i965_encoder.c
QuPengfei: move it into the function gen9_mfc_context_init and gen9_vme_context_init

src/gen9_avc_encoder.c
```diff
@@ -5748,6 +5748,1596 @@ gen9_avc_kernel_init(VADriverContextP ctx,
...
+                          0);
+        } else {
+            OUT_BCS_BATCH(batch, 0);
+            OUT_BCS_BATCH(batch, 0);
+        }
+    }
+
+    OUT_BCS_BATCH(batch, i965->intel.mocs_state);
+
+    /* the DW34-36 is the MV for the current reference */
```
QuPengfei: that is correct. it will be fixed.

src/gen9_avc_encoder.c
```diff
@@ -5748,6 +5748,1596 @@ gen9_avc_kernel_init(VADriverContextP ctx,
...
+gen9_mfc_avc_qm_state(VADriverContextP ctx,
+                      struct encode_state *encode_state,
+                      struct intel_encoder_context *encoder_context)
+{
+    struct encoder_vme_mfc_context * pak_context = (struct encoder_vme_mfc_context *)encoder_context->vme_context;
+    struct avc_enc_state * avc_state = (struct avc_enc_state * )pak_context->private_enc_state;
+    VAEncSequenceParameterBufferH264  *seq_param = avc_state->seq_param;
+    VAEncPictureParameterBufferH264  *pic_param = avc_state->pic_param;
+
+    /* TODO: add support for non flat matrix */
```
QuPengfei: yes.the comments will be removed.

src/gen9_avc_encoder.c
```diff
@@ -5748,6 +5748,1596 @@ gen9_avc_kernel_init(VADriverContextP ctx,
...
+                    avc_state->rounding_value = gen9_avc_inter_rounding_b[generic_state->preset];
+            }else
+            {
+                avc_state->rounding_value = avc_state->rounding_inter_b;
+            }
+        }
+    }
+
+    slice_hor_pos = slice_param->macroblock_address % generic_state->frame_width_in_mbs;
+    slice_ver_pos = slice_param->macroblock_address / generic_state->frame_height_in_mbs;
```
QuPengfei: it will be fixed

src/gen9_avc_encoder.c
```diff
@@ -5748,6 +5748,1596 @@ gen9_avc_kernel_init(VADriverContextP ctx,
...
+
+    if (next_slice_param) {
+        next_slice_hor_pos = next_slice_param->macroblock_address % generic_state->frame_width_in_mbs;
+        next_slice_ver_pos = next_slice_param->macroblock_address / generic_state->frame_height_in_mbs;
```
QuPengfei: it will be fixed

src/gen9_avc_encoder.c
QuPengfei: these info also help debug. i suggest keep them.

src/gen9_avc_encoder.c
```diff
@@ -5748,6 +5748,1596 @@ gen9_avc_kernel_init(VADriverContextP ctx,
...
+            slice_param++;
+        }
+    }
+    avc_state->enable_avc_ildb = enable_avc_ildb;
+
+    /* setup the all surface and buffer for PAK */
+    /* Setup current reconstruct frame */
+    obj_surface = encode_state->reconstructed_object;
+    va_status = i965_check_alloc_surface_bo(ctx, obj_surface, 1, VA_FOURCC_NV12, SUBSAMPLE_YUV420);
+
```
QuPengfei: here i965_check_alloc_surface_bo will be removed.

src/gen9_avc_encoder.c
```diff
@@ -0,0 +1,6031 @@
...
+    /* TODO: set paramters DW19/DW20 for slices */
+}
+
+void gen9_avc_set_image_state(VADriverContextP ctx,
+                              struct encode_state *encode_state,
+                              struct intel_encoder_context *encoder_context,
+                              struct i965_gpe_resource *gpe_resource)
+{
+    struct encoder_vme_mfc_context * pak_context = (struct encoder_vme_mfc_context *)encoder_context->vme_context;
+    struct generic_enc_codec_state * generic_state = (struct generic_enc_codec_state * )pak_context-
```
xhaihao: free() can accept NULL pointer

xhaihao: free() can accept NULL pointer

src/gen9_avc_encoder.c
```diff
@@ -5748,6 +5748,1596 @@ gen9_avc_kernel_init(VADriverContextP ctx,
...
+    struct gen9_avc_encoder_context * avc_ctx = (struct gen9_avc_encoder_context * )pak_context->private_enc_ctx;
+    struct intel_batchbuffer *batch = encoder_context->base.batch;
+    int i;
+
+    BEGIN_BCS_BATCH(batch, 65);
+
+    OUT_BCS_BATCH(batch, MFX_PIPE_BUF_ADDR_STATE | (65 - 2));
+
+    /* the DW1-3 is for pre_deblocking */
+    OUT_BUFFER_3DW(batch, avc_ctx->res_pre_deblocking_output.bo, 1, 0, 0);
```
QuPengfei: yes,  MOCS setting is used in default for OUT_BUFFER_3DW.

src/i965_encoder_common.c
```diff
@@ -0,0 +1,124 @@
...
+            0x120FF10F, 0x1E22E20D, 0x20E2FF10, 0x2EDD06FC, 0x11D33FF1, 0xEB1FF33D, 0x4EF1F1F1, 0xF1F21211,
+            0x0DFFFFE0, 0x11201F1F, 0x1105F1CF, 0x00000000, 0x00000000, 0x00000000, 0x00000000, 0x00000000
+        },
+        // MEMethod: 7
+        {
+            0x1F11F10F, 0x2E22E2FE, 0x20E220DF, 0x2EDD06FC, 0x11D33FF1, 0xEB1FF33D, 0x02F1F1F1, 0x1F201111,
+            0xF1EFFF0C, 0xF01104F1, 0x10FF0A50, 0x000FF1C0, 0x00000000, 0x00000000, 0x00000000, 0x00000000
+        }
+    }
+};
```
QuPengfei: i965_encoder.c is mainly for generic functions implementation and  i965_encoder_common.c is for common table defination. We can put all the encoder generic tables here. And it make more clear and easy to read.

src/i965_encoder_common.h
QuPengfei: yes, i can move these two structure into gen9_avc_encoder.h And refine it later.

src/gen9_avc_encoder.c
```diff
@@ -0,0 +1,6031 @@
...
+gen9_avc_update_misc_parameters(VADriverContextP ctx,
+                                struct encode_state *encode_state,
+                                struct intel_encoder_context *encoder_context)
+{
+    struct encoder_vme_mfc_context * vme_context = (struct encoder_vme_mfc_context *)encoder_context->vme_context;
+    struct generic_enc_codec_state * generic_state = (struct generic_enc_codec_state * )vme_context->generic_enc_state;
+    int i;
+
+    /* brc */
+    generic_state->max_bit_rate = ALIGN(encoder_context->brc.bits_per_second[0], 1000) / 1000;
```
QuPengfei: it will be fixed

src/gen9_avc_encoder.c
```diff
@@ -0,0 +1,6031 @@
...
+            generic_state->brc_need_reset = 1;
+        }
+    }
+
+    /*  frame rate */
+    if (generic_state->internal_rate_mode != VA_RC_CQP)
+    {
+        generic_state->frames_per_100s = encoder_context->brc.framerate[0].num * 100/encoder_context->brc.framerate[0].den ;
+        generic_state->frame_rate = encoder_context->brc.framerate[0].num/encoder_context->brc.framerate[0].den ;
+        generic_state->window_size = (int)(encoder_context->brc.window_size * generic_state->frame_rate /1000);// brc.windows size in ms as the unit
```
QuPengfei: the variable name will be change to frames_per_window_size

src/gen9_avc_encoder.c
```diff
@@ -0,0 +1,6031 @@
...
+    unsigned int frame_mb_nums = frame_width_in_mbs * frame_height_in_mbs;
+    int allocate_flag = 1;
+    int width,height;
+
+    if (!obj_surface || !obj_surface->bo)
+        return VA_STATUS_ERROR_INVALID_SURFACE;
+
+    if (obj_surface->private_data &&
+        obj_surface->free_private_data != gen9_free_surfaces_avc) {
+        obj_surface->free_private_data(&obj_surface->private_data);
```
QuPengfei: According to mail discussion, it is a common issue for all encoder.it should file a bug and fix it in seperate patch.

src/gen9_avc_encoder.c
```diff
@@ -0,0 +1,6031 @@
...
+    intel_batchbuffer_start_atomic(batch, 0x1000);
+
+    status_buffer = &(avc_ctx->status_buffer);
+    memset(&mi_store_data_imm, 0, sizeof(mi_store_data_imm));
+    mi_store_data_imm.bo = status_buffer->bo;
+    mi_store_data_imm.offset = status_buffer->media_index_offset;
+    mi_store_data_imm.dw0 = media_function;
+    gen8_gpe_mi_store_data_imm(ctx, batch, &mi_store_data_imm);
+
+    intel_batchbuffer_emit_mi_flush(batch);
```
QuPengfei: sure, it will be fixed

src/gen9_avc_encoder.c
QuPengfei: sure. the free() will be call directly.

QuPengfei: sure. the free() will be call directly.

src/gen9_avc_encoder.c
xhaihao: Yes, we should fix private data usage in another patch. What I mean is not to release private data here because releasing private data will result in other issues.  BTW 
fhvwy has created https://github.com/01org/intel-vaapi-driver/issues/42 to track this issue

src/i965_encoder_common.c
xhaihao: It would be better to use a specific file name for constant data, like as the file used for AVC constant (gen9_avc_const_def.c)

src/i965_gpe_utils.c
QuPengfei: it will be update accordingly

seanvk: Agree, other alignments can be set as needed.

src/i965_encoder_common.c
seanvk: Agree, let's favor more specific titles for the file.  Saying common implies we are going to move more tables here, which we're not currently doing.

src/i965_encoder_common.h
seanvk: Yes, this goes along with the concern about the constants in the common file too.  If we aren't doing it now, let's keep it specific until we are ready to make the change in a later patch series.

src/Makefile.am
```diff
@@ -161,6 +161,7 @@ source_h = \
         i965_encoder_common.h        \
         i965_encoder_api.h        \
         gen9_avc_const_def.h      \
+        gen9_avc_encoder_kernels.h      \
```
seanvk: The kernels should be using the APACHE 2 license and NOT the MIT license.  Please see VP8 patch series comments and subsequent changes by @xhaihao .  Same rule applies for AVC kernels.

src/Makefile.am
```diff
@@ -104,6 +104,7 @@ source_c = \
 	intel_common_vpp_internal.c           \
         i965_encoder_common.c    \
         gen9_avc_const_def.c    \
+        i965_avc_encoder_common.c \
```
seanvk: Change to AVC in name

src/Makefile.am
```diff
@@ -162,6 +163,7 @@ source_h = \
         i965_encoder_api.h        \
         gen9_avc_const_def.h      \
         gen9_avc_encoder_kernels.h      \
+        i965_avc_encoder_common.h       \
```
seanvk: Change to AVC in name

src/gen9_avc_encoder.c
seanvk: @xhaihao You know this issue has caught people more than once.  In future we may need to explicitly define ALIGN to be something ALIGNPWR2 or something.

src/gen9_avc_encoder.c
seanvk: agree

src/gen9_avc_encoder.c
seanvk: Per discussion on VP8, we should at a minimum make a change in this patch series such that private_data is released if the private_data is not expected.  We can later remove it once we have a common solution for private data usage. @yakuizhao

src/gen9_avc_encoder.c
seanvk: Please do make sure elsewhere in the code you are flushing before you do the store, not just at this location.

src/gen9_avc_encoder.c
```diff
@@ -5748,6 +5748,1596 @@ gen9_avc_kernel_init(VADriverContextP ctx,
...
+            luma_log2_weight_denom = 5;
+            chroma_log2_weight_denom = 5;
+        }
+    }
+
+    max_qp_n = 0;       /* TODO: update it */
+    max_qp_p = 0;       /* TODO: update it */
+    grow = 0;           /* TODO: update it */
+    shrink = 0;         /* TODO: update it */
+
```
seanvk: I prefer not to see "TODO" in the code.  Rather, you should create a Github issue and assign it to yourself detailing the additional changes needed.  @QuPengfei

src/gen9_avc_encoder.c
seanvk: I agree with @QuPengfei, although they are not used, they are informative and saves time from having to jump to the PRM for the DW definitions.  No harm is done and makes the code easier to maintain.

src/i965_drv_video.c
seanvk: I'm assuming you are including KBL compatible kernels in the patch?

src/i965_drv_video.h
seanvk: Yes, these kernels do support 1-7 and we need to support them for customer use cases.

src/i965_encoder.c
seanvk: Yes, good catch.

src/i965_encoder_common.c
QuPengfei: Yes, rename the file as i965_encoder_const_def.c

src/gen9_avc_encoder.c
```diff
@@ -5748,6 +5748,1596 @@ gen9_avc_kernel_init(VADriverContextP ctx,
...
+            OUT_BCS_BATCH(batch, 0);
+            OUT_BCS_BATCH(batch, 0);
+        }
+    }
+
+    OUT_BCS_BATCH(batch, i965->intel.mocs_state);
+
+    /* the DW34-36 is the MV for the current reference */
+    OUT_BCS_RELOC64(batch, avc_ctx->res_direct_mv_buffersr[NUM_MFC_AVC_DMV_BUFFERS - 2].bo,
+                  I915_GEM_DOMAIN_INSTRUCTION, 0,
```
seanvk: Good catch.  We should make sure we have all the right domains set.  It's not an uncommon problem.

src/Makefile.am
QuPengfei: license header will be changed to APACHE2 accordingly for kernel file.

src/gen9_avc_encoder.c
QuPengfei: the comments will removed. by now  no addational changes needed.

src/gen9_avc_encoder.c
xhaihao: Actually reading/writing too many unused registers and memories in the pipeline will impact the performance a little bit.

src/gen9_avc_encoder.c
QuPengfei: agree. more locations should be checked.

QuPengfei:
update the patch. Please help to review.

src/gen9_avc_encoder.c
```diff
@@ -5736,6 +5736,1586 @@ gen9_avc_kernel_init(VADriverContextP ctx,
...
+        avc_state->slice_batch_offset[i] = 0;
+    }
+
+
+    size = w_mb * 64;
+    i965_free_gpe_resource(&avc_ctx->res_intra_row_store_scratch_buffer);
+    allocate_flag = i965_allocate_gpe_resource(i965->intel.bufmgr,
+                                 &avc_ctx->res_intra_row_store_scratch_buffer,
+                                 size,
+
```
yakuizhao: Please free the pres_slice_batch_buffer_2nd_level before submitting the batch buffer.
Otherwise the buffer domain of pres_slice_batch_buffer_2nd_level is incorrect.

src/i965_gpe_utils.c
```diff
@@ -1385,14 +1385,15 @@ i965_allocate_gpe_resource(dri_bufmgr *bufmgr,
...
+                                       struct object_surface *obj_surface,
+                                       unsigned int alignment)
 {
     unsigned int swizzle;
 
     res->type = I965_GPE_RESOURCE_2D;
-    res->width = obj_surface->orig_width;
-    res->height = obj_surface->orig_height;
+    res->width = ALIGN(obj_surface->orig_width,alignment);
+    res->height = ALIGN(obj_surface->orig_height,alignment);
```
pengche1: is it a mistake?

src/i965_avc_encoder_common.h
```diff
@@ -0,0 +1,300 @@
...
+    INTEL_AVC_LEVEL_42                   = 42,
+    INTEL_AVC_LEVEL_5                    = 50,
+    INTEL_AVC_LEVEL_51                   = 51,
+    INTEL_AVC_LEVEL_52                   = 52
+} INTEL_AVC_LEVEL_IDC;
+
+struct gen9_mfx_avc_img_state
+{
+    union {
+        struct {
```
pengche1: is it the common structure or special for gen9?

src/gen9_avc_encoder.c
```diff
@@ -5746,245 +5749,1835 @@ gen9_avc_kernel_init(VADriverContextP ctx,
...
+            }
+        }
+    }
+
+    slice_hor_pos = slice_param->macroblock_address % generic_state->frame_width_in_mbs;
+    slice_ver_pos = slice_param->macroblock_address / generic_state->frame_height_in_mbs;
+
+    if (next_slice_param) {
+        next_slice_hor_pos = next_slice_param->macroblock_address % generic_state->frame_width_in_mbs;
+        next_slice_ver_pos = next_slice_param->macroblock_address / generic_state->frame_height_in_mbs;
```
QuPengfei: it is already fixed in this series of patches.

src/gen9_avc_const_def.h
```diff
@@ -0,0 +1,115 @@
...
+extern const unsigned char gen9_avc_max_ref_id0_progressive_4k[PRESET_NUM];
+extern const unsigned char gen9_avc_max_ref_id0[PRESET_NUM];
+extern const unsigned char gen9_avc_max_b_ref_id0[PRESET_NUM];
+extern const unsigned char gen9_avc_max_ref_id1[PRESET_NUM];
+extern const unsigned int gen9_avc_inter_rounding_p[PRESET_NUM];
+extern const unsigned int gen9_avc_inter_rounding_b_ref[PRESET_NUM];
+extern const unsigned int gen9_avc_inter_rounding_b[PRESET_NUM];
+// This applies only for progressive pictures. For interlaced, CAF is currently not disabled.
+extern const unsigned int gen9_avc_disable_all_fractional_check_for_high_res[PRESET_NUM];
+extern unsigned char gen9_avc_adaptive_inter_rounding_p[AVC_QP_MAX];
```
QuPengfei: sure, it will be fixed in the next version

src/i965_gpe_utils.c
QuPengfei: no, this function is re defined with input parameter alignment

src/gen9_avc_encoder.c
```diff
@@ -0,0 +1,6008 @@
...
+    0x10001000, 0x10001000, 0x10001000, 0x10001000,
+    0x10001000, 0x10001000, 0x10001000, 0x10001000,
+    0x10001000, 0x10001000, 0x10001000, 0x10001000,
+    0x10001000, 0x10001000, 0x10001000, 0x10001000,
+    0x10001000, 0x10001000, 0x10001000, 0x10001000,
+    0x10001000, 0x10001000, 0x10001000, 0x10001000,
+    0x10001000, 0x10001000, 0x10001000, 0x10001000
+};
+
+static unsigned int slice_type_kernel[3] = {1,2,0};
```
QuPengfei: it will be fixed in the next version

src/i965_encoder_common.h
```diff
@@ -0,0 +1,517 @@
...
+        VADriverContextP ctx,
+        struct encode_state *encode_state,
+        struct i965_gpe_context *gpe_context,
+        struct intel_encoder_context *encoder_context,
+        void *param);
+};
+/*
+   the definition for encoder codec state
+*/
+
```
QuPengfei: for fame_width_in_pixel and frame_height_in_pixel and etc, these variabls are used almost in each function. it is easy to acces and read to put them together in one structure. I think it is OK.

for frame_width_in_mbs, it is common for AVC/VP8, and even if HEVC/VP9, it is also used on some caculation when MB alignment. 
frame_width_in_ctb/cb/cu/lcu is specified for HEVC only,VP should use SB related. that is the reason there are no definition in this structure generic_enc_codec_state here.

src/gen9_avc_encoder.c
QuPengfei: it is confused. how about the first level batch buffer? Does it need the same free operation before the batch buffer submitting? or Only second level batch buffer need this operation?

src/i965_avc_encoder_common.h
QuPengfei: it is special for gen9. i will move it to gen9_avc_encoder.h

src/gen9_avc_encoder.c
```diff
@@ -5736,6 +5736,1586 @@ gen9_avc_kernel_init(VADriverContextP ctx,
...
+
+    /* the DW16-18 is for the deblocking filter */
+    OUT_BUFFER_3DW(batch, avc_ctx->res_deblocking_filter_row_store_scratch_buffer.bo, 1, 0, i965->intel.mocs_state);
+
+    /* the DW 19-50 is for Reference pictures*/
+    for (i = 0; i < ARRAY_ELEMS(avc_ctx->list_reference_res); i++) {
+        OUT_BUFFER_2DW(batch, avc_ctx->list_reference_res[i].bo, 1, 0);
+    }
+
+    /* DW 51, reference picture attributes */
```
QuPengfei: sure, this will be fixed in the next version

src/gen9_avc_encoder.c
QuPengfei: yes, it is clear for me. it will be fixed.

QuPengfei: yes, it will be fixed.

seanvk:
@QuPengfei When will we see an updated PR from you?  I'll like to see this soon.  I know you have a lot of feedback.

seanvk:
@QuPengfei  your commit headers are missing:

Fixes #n 

I'd like to see you add those so the patches are actually linked to the issues for QA.

Thanks,

QuPengfei:
update the patch according to comments again. Please review.
Note:
1. add Fix # in the commit message
2. fix the second level batch buffer issue, which will free before submitting the command.
3. misc changes.

src/gen9_avc_encoder.c
```diff
@@ -0,0 +1,6008 @@
...
+    0x10001000, 0x10001000, 0x10001000, 0x10001000,
+    0x10001000, 0x10001000, 0x10001000, 0x10001000,
+    0x10001000, 0x10001000, 0x10001000, 0x10001000,
+    0x10001000, 0x10001000, 0x10001000, 0x10001000,
+    0x10001000, 0x10001000, 0x10001000, 0x10001000
+};
+
+static const unsigned int slice_type_kernel[3] = {1,2,0};
+
+const gen9_avc_brc_init_reset_curbe_data gen9_avc_brc_init_reset_curbe_init_data =
```
yakuizhao: Please add the "static" prefix for the corresponding *_init_data to keep the unique code-style.

QuPengfei: yes, i will fix it

QuPengfei:
update the patch according to comments again. Please review.
Note:
1. fix CQP issue
2.  add static before const table.

QuPengfei:
I rebase the patch to the latest and update the commit header. by now there is no conflict with master.

src/gen9_avc_encoder_kernels.h
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+ * OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT.
+ * IN NO EVENT SHALL PRECISION INSIGHT AND/OR ITS SUPPLIERS BE LIABLE FOR
+ * ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+ * TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ *
+ * This file was originally licensed under the following license
```
seanvk: We are not relicensing the file. You must retain only the Apache 2 license in the header. Remove the MIT license and leave only the Apache license in this file.  We cannot change the license of the Media Kernel team. @QuPengfei

QuPengfei: sure, I will correct it.

BTW, I notice the VP8 use the same MIT+Apache header for the binary. I think it has the same issue.

src/gen9_avc_encoder.c
```diff
@@ -5736,6 +5736,1596 @@ gen9_avc_kernel_init(VADriverContextP ctx,
...
+                                 struct intel_batchbuffer *batch)
+{
+    struct encoder_vme_mfc_context * pak_context = (struct encoder_vme_mfc_context *)encoder_context->vme_context;
+    struct avc_enc_state * avc_state = (struct avc_enc_state * )pak_context->private_enc_state;
+    VAPictureH264 *ref_pic;
+    int i, slice_type, ref_idx_shift;
+    unsigned int fwd_ref_entry;
+    unsigned int bwd_ref_entry;
+
+    /* max 4 ref frames are allowed for l0 and l1 */
```
yakuizhao: It seems that the code of gen9_mfc_avc_ref_idx_state looks too hacky. Why is it limited to 4?
It will be better to be more generic.

src/gen9_avc_encoder.c
```diff
@@ -0,0 +1,6008 @@
...
+                              struct i965_gpe_resource *gpe_resource)
+{
+    struct encoder_vme_mfc_context * pak_context = (struct encoder_vme_mfc_context *)encoder_context->vme_context;
+    struct generic_enc_codec_state * generic_state = (struct generic_enc_codec_state * )pak_context->generic_enc_state;
+    char *pdata;
+    int i;
+    unsigned int * data;
+    struct gen9_mfx_avc_img_state cmd;
+
+    pdata = i965_map_gpe_resource(gpe_resource);
```
yakuizhao: Is the frame_store_id really used?
If not , please remove it.

src/gen9_avc_encoder.c
QuPengfei: For AVC Spec, Max reference frame is 4 for 1080P and 9 for 720P and 16 for 720x480.

Generally 4 ref frame is enough and no quality improvement even more ref frames.

src/gen9_avc_encoder.c
yakuizhao: Please remove the VDENC as this patch doesn't handle the vdenc.

src/gen9_avc_encoder.c
yakuizhao: Although 4 is enough for the mentioned scenario,  the code still can be more generic. For example: when the PAK mode is used and user provides the corresponding PAK_OBJ cmd list, the code is still appropriate.

src/gen9_avc_encoder.c
QuPengfei: there is no the VDENC code here. it is the comment for non-vdenc command field.

src/gen9_avc_encoder.c
yakuizhao: From the H264 spec it seems that the max num of reference_frames can be 16 for the 1080p@5.1 Level. 
At the same time based on the code it seems that it is only mentioned in comemnt while it is not checked in the code.

src/gen9_avc_encoder.c
```diff
@@ -5736,6 +5736,1596 @@ gen9_avc_kernel_init(VADriverContextP ctx,
...
+    OUT_BUFFER_3DW(batch, avc_ctx->res_pak_mb_status_buffer.bo, 1, 0, i965->intel.mocs_state);//?
+
+    /* the DW13-15 is for the intra_row_store_scratch */
+    OUT_BUFFER_3DW(batch, avc_ctx->res_intra_row_store_scratch_buffer.bo, 1, 0, i965->intel.mocs_state);
+
+    /* the DW16-18 is for the deblocking filter */
+    OUT_BUFFER_3DW(batch, avc_ctx->res_deblocking_filter_row_store_scratch_buffer.bo, 1, 0, i965->intel.mocs_state);
+
+    /* the DW 19-50 is for Reference pictures*/
+    for (i = 0; i < ARRAY_ELEMS(avc_ctx->list_reference_res); i++) {
```
yakuizhao: Is the write meaningful?
If not, please remove it.

src/gen9_avc_encoder.c
QuPengfei: correct, I will add the check in the code.

src/gen9_avc_encoder.c
QuPengfei: yes, it store the frame id

src/gen9_avc_encoder.c
QuPengfei: for reference surface, there is only read only operation. it will be removed.

QuPengfei:
update the patch, please help to review.

Note:
1. change license header of the kernel binary to Apache 2 only
2. add the check for Max ref frame number
3. set the MOCS to zero for ref surface

src/gen9_avc_const_def.c
```diff
@@ -0,0 +1,1090 @@
...
+ * copy of this software and associated documentation files (the
+ * "Software"), to deal in the Software without restriction, including
+ * without limitation the rights to use, copy, modify, merge, publish,
+ * distribute, sub license, and/or sell copies of the Software, and to
+ * permit persons to whom the Software is furnished to do so, subject to
+ * the following conditions:
+ *
+ * The above copyright notice and this permission notice (including the
+ * next paragraph) shall be included in all copies or substantial portions
+ * of the Software.
```
pengche1: should we also change it to alpha?

pengche1: should we also change it to apache license?

src/i965_avc_encoder_common.h
```diff
@@ -0,0 +1,321 @@
...
+    INTEL_AVC_LEVEL_5                    = 50,
+    INTEL_AVC_LEVEL_51                   = 51,
+    INTEL_AVC_LEVEL_52                   = 52
+} INTEL_AVC_LEVEL_IDC;
+
+/*
+common structure and define
+*/
+struct gen9_avc_encoder_context {
+
```
pengche1: This context definition is special for GEN9. why not move it to gen9_avc_encoder.h?

src/i965_avc_encoder_common.h
```diff
@@ -0,0 +1,321 @@
...
+    struct gen_avc_wp_context  context_wp;
+    struct gen_avc_sfd_context  context_sfd;
+
+    struct encoder_status_buffer_internal status_buffer;
+
+};
+
+#define MAX_AVC_SLICE_NUM 256
+struct avc_enc_state {
+
```
pengche1: This context definition is special for GEN9. why not move it to gen9_avc_encoder.h?

QuPengfei: it is common for AVC on GenX. I am thinking about supporting AVC on more platform, such as KBL/BDW etc and move this structure into the common.

src/gen9_avc_encoder.c
```diff
@@ -5736,6 +5736,1596 @@ gen9_avc_kernel_init(VADriverContextP ctx,
...
+    unsigned int fwd_ref_entry;
+    unsigned int bwd_ref_entry;
+
+    /* max 4 ref frames are allowed for l0 and l1 */
+    fwd_ref_entry = 0x80808080;
+    slice_type = intel_avc_enc_slice_type_fixup(slice_param->slice_type);
+
+    if ((slice_type == SLICE_TYPE_P) ||
+        (slice_type == SLICE_TYPE_B)) {
+          for (i = 0; i < MAX(avc_state->num_refs[0],4); i++) {
```
yakuizhao: It seems that the min(avc_state->num_refs, 4) should be used instead of MAX.
Please fix it.

src/i965_avc_encoder_common.h
QuPengfei: I am thinking about supporting AVC on more platform, such as KBL/BDW etc and move this structure into the common. Yes, the structure name is gen9, i will change it to i965_avc_encoder_context to be common for other platforms.

src/gen9_avc_encoder.c
QuPengfei: yes, i will fix it.

src/gen9_avc_encoder.c
```diff
@@ -0,0 +1,7615 @@
...
+    allocate_flag = i965_allocate_gpe_resource(i965->intel.bufmgr,
+                             &avc_ctx->res_image_state_batch_buffer_2nd_level,
+                             ALIGN(size,0x1000),
+                             "second levle batch (image state write) buffer");
+    if (!allocate_flag)
+        goto failed_allocation;
+
+    /* scaling related surface   */
+    if(avc_state->mb_status_supported)
+    {
```
pengche1: minor suggestion. keep the identical coding type as others. like this
if () {

}

src/gen9_avc_encoder.c
```diff
@@ -0,0 +1,7615 @@
...
+
+    generic_ctx->pfn_send_scaling_surface(ctx,encode_state,gpe_context,encoder_context,&surface_param);
+
+
+    gen8_gpe_setup_interface_data(ctx, gpe_context);
+
+    memset(&kernel_walker_param, 0, sizeof(kernel_walker_param));
+    if(surface_param.use_32x_scaling)
+    {
+        kernel_walker_param.resolution_x = downscaled_width_in_mb ;
```
pengche1: delete the blank space before ;

src/gen9_avc_encoder.c
```diff
@@ -0,0 +1,7615 @@
...
+                                     width, height,
+                                     width,
+                                     "external mb qp buffer");
+        if (!allocate_flag)
+            goto failed_allocation;
+    }
+
+
+    /* maybe it is not needed by now. it is used in crypt mode*/
+    i965_free_gpe_resource(&avc_ctx->res_brc_mbenc_curbe_write_buffer);
```
pengche1: I suggest to delete this comment or remove this buffer temporarily

src/gen9_avc_encoder.c
QuPengfei: agree. keep the same style

src/gen9_avc_encoder.c
QuPengfei: agree. remove it.

src/gen9_avc_encoder.c
QuPengfei: agree.

src/gen9_avc_encoder.c
QuPengfei: agree. keep the same style as the following
if()
{

}

src/gen9_avc_encoder.c
```diff
@@ -0,0 +1,7615 @@
...
+        return VA_STATUS_ERROR_ALLOCATION_FAILED;
+    }
+
+    i965_check_alloc_surface_bo(ctx, avc_surface->scaled_16x_surface_obj, 1,
+                                VA_FOURCC('N', 'V', '1', '2'), SUBSAMPLE_YUV420);
+
+    downscaled_width_32x = generic_state->frame_width_32x;
+    downscaled_height_32x = generic_state->frame_height_32x;
+    i965_CreateSurfaces(ctx,
+                        downscaled_width_32x,
```
QuPengfei: agree. allocate the 32x surface according to the HME flag

src/gen9_avc_const_def.c
QuPengfei: no,apache2 only for media kernel binary

QuPengfei:
update the patch. Please review.

Note:
1. allocate the 32x surface according to the HME flag
2. remove the surface res_brc_mbenc_curbe_write_buffer
3. fix ref num checking
4. change the structure name gen9_avc_encoder_context to i965_avc_encoder_context 
4. misc changes

QuPengfei:
rebase to master.

note:
1. fix the surface read/write domain for read only operation

xhaihao:
I see a strange flag of <U+FEFF> in some new files in your commits, could you double check each commit and try to remove this flag?

For example, 
git log -p ef9cc76747afa4c29c211972971b9863a1e58c9e

diff --git a/src/gen9_avc_encoder.h b/src/gen9_avc_encoder.h
new file mode 100644
index 0000000..a00b058
--- /dev/null
+++ b/src/gen9_avc_encoder.h
@@ -0,0 +1,2345 @@
+<U+FEFF>/*
+ * Copyright @ 2016 Intel Corporation

xhaihao:
Please change the year of 2016 to 2017 in the copyright statement in a new file.

QuPengfei:
update the patch.

Note:
1.  change the license header in the new file "2016" to "2017.
2. remove the strange symbal in new file

PR #34, opened 2017-02-08
JPEG Encoder: support vaQuerySurfaceAttributes() pixel format

ceyusa:
When querying the surface attributes, using the JPEG encoder
(profile VAProfileJPEGBaseline, entrypoint VAEntrypointEncPicture)
the function doesn't return the available VASurfaceAttribPixelFormat
because it is not considered in the driver's code.

This patch proposes to return the default branch for encoders entry
points.

Signed-off-by: Víctor Manuel Jáquez Leal <vjaquez@igalia.com>

### DIFF
--- a/src/i965_drv_video.c
+++ b/src/i965_drv_video.c
@@ -5845,7 +5845,8 @@ i965_QuerySurfaceAttributes(VADriverContextP ctx,
             }
         } else if (obj_config->entrypoint == VAEntrypointEncSlice ||  /* encode */
                    obj_config->entrypoint == VAEntrypointVideoProc ||
-                   obj_config->entrypoint == VAEntrypointEncSliceLP) {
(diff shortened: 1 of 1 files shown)


ceyusa:
This patch comes from this bug report in gstreamer-vaapi
https://bugzilla.gnome.org/show_bug.cgi?id=775490

seanvk:
Also this issue should have been caught in libva-utils gtests but due to 01org/libva-utils#1 it was skipped.

seanvk:
Hmm, the test is skipping everywhere for VAEntrypointEncPicture which is wrong on KBL.  The tests definitely need to be fixed in 01org/libva-utils#1

ceyusa:
I have pushed a new commit with the requested commit log, associated with the issue.

uartie:
This actually seems like a candidate for a test in the intel driver gtests, not the vaapi gtests.

PR #49, opened 2017-02-14
configure: check for x11 on path

### LINKED ISSUE #48: Properly check for X11 backend to avoid compilation break
configuring should use pkg-config to know if X11 headers (distributions call this devel packages) are available to the path and avoid compiling X11 stuff if they are not present.   

Currently configure won't check for it and as a result in the absence of X11 headers the build breaks when compiling vi_dricommon.h, which is included by va_backend_compat.h.  Compilation relies on a successful configure stage.

chivakker:
enable-x11 is set to yes by default and during configuration it has to
check for its header files on the path before breaking the compilation
if those are not present

va_dricommon.h requires x11 on path to compile

Fixes #48


configure.ac
```diff
@@ -141,12 +141,19 @@ AC_DEFINE_UNQUOTED([VA_DRIVER_INIT_FUNC], [$VA_DRIVER_INIT_FUNC],
...
 dnl Check for VA/DRM API
-USE_X11="$enable_x11"
-if test "$USE_X11" = "yes"; then
-    PKG_CHECK_MODULES(LIBVA_X11_DEPS, [libva-x11],
-      [AC_DEFINE([HAVE_VA_X11], [1], [Defined to 1 if VA/X11 API is enabled])],
-      [USE_X11="no"])
+USE_X11="no"
+if test "$enable_x11" = "yes"; then
+    USE_X11="yes"
+    PKG_CHECK_MODULES([X11],    [x11],    [:], [])
```
seanvk: If you are going to do this for X11, you might as well do the same for libva-drm which needs to have drm headers support.  That way you are somewhat consistent with Wayland, egl, etc...

chivakker: Good point, I'll propose a fix that will cover those combinations

configure.ac
```diff
@@ -141,12 +141,19 @@ AC_DEFINE_UNQUOTED([VA_DRIVER_INIT_FUNC], [$VA_DRIVER_INIT_FUNC],
     [Define driver entry-point])
 
 dnl Check for VA/DRM API
-USE_X11="$enable_x11"
-if test "$USE_X11" = "yes"; then
-    PKG_CHECK_MODULES(LIBVA_X11_DEPS, [libva-x11],
-      [AC_DEFINE([HAVE_VA_X11], [1], [Defined to 1 if VA/X11 API is enabled])],
-      [USE_X11="no"])
+USE_X11="no"
+if test "$enable_x11" = "yes"; then
```
chivakker: libva has those checks as sample applications needed them.  As they get moved to libva-utils, libva doesn't need it anymore and instead libva-utils should also cover it.  Patches to all projects will be submitted.

yakuizhao: @chivakker maybe the similar X11 check in libva is also added for the driver.
>X11/XEXT/XFIXES  check is added.

seanvk: @chivakker Libva is **still** responsible for the backend, regardless of the move of the test folder, deps are still there.  Let's focus on the issue here.  What is the problem that you are trying to solve specifically for this project.

xhaihao: @seanvk, a user should install the develop package for X11, otherwise it will break the compilation although user installs the develop packages for libva.

seanvk: @xhaihao that is not what I'm talking about. Of course drivel header deps exist. My point is the dross are still needed regardless of the test folder move.

seanvk: I'm going to start requiring a how to reproduce line and configuration in issues. Finally commits will need a verify/test line in the commit header just like chromium. That avoids this thrashing

seanvk:
@yakuizhao Please update your review

seanvk:
@xhaihao please update your review

yakuizhao:
It will be better that the similar check in libva is added for the driver.
   >  PKG_CHECK_MODULES([X11],    [x11],    [:], [USE_X11="no"])
    PKG_CHECK_MODULES([XEXT],   [xext],   [:], [USE_X11="no"])
    PKG_CHECK_MODULES([XFIXES], [xfixes], [:], [USE_X11="no"])
 
Then the LIBVA_X11 is checked.

Thanks

seanvk:
I tend to agree with @yakuizhao, we should error on the side of simply doing the PKG check.  We should not over-guess.  Simply look for deps through pkgconfig and if they are not there then fail.

chivakker:
Sorry I didn't reply earlier.  

I need to take it one or two steps back.

This is the scenario (let's assume that X11.h

$ cd libva
$ autogen.sh --prefix=<my_prefix> --enable-x11
...
checking for X11... no
checking for XEXT... no
checking for XFIXES... no
checking for EGL... no
...
libva - 1.8.0.pre1 (VA-API 0.40.0)

Installation prefix .............. : <my_prefix>
Default driver path .............. : ${exec_prefix}/lib/dri
Extra window systems ............. : drm
Build dummy driver ............... : yes
Build documentation .............. : no
Build with messaging ............. : yes
...
$make install
$cd intel-vaapi-driver
$ autogen.sh --prefix=<my_prefix> --enable-x11
...
checking for LIBVA_DEPS... yes
checking for LIBVA_DRM_DEPS... yes
checking drm_fourcc.h usability... yes
checking drm_fourcc.h presence... yes
checking for drm_fourcc.h... yes
checking for LIBVA_X11_DEPS... yes
...
intel_vaapi_driver configuration summary:

VA-API version ................... : 0.40.0
VA-API drivers path .............. : /home/dcharles/lib/dri
Windowing systems ................ : drm x11
Build tests ...................... : no
...
$make -j5
...
In file included from i965_output_dri.c:27:0:
<my_prefix>/include/va/va_dricommon.h:28:22: fatal error: X11/Xlib.h: No such file or directory
 #include <X11/Xlib.h>
...

Observations:

libva compiles file as it finds x11 pkg_config information is not available on the system setup.  Then intel-vaapi-driver assumes x11 is available, i.e. LIBVA_X11_DEPS... yes

Then the build is broken due to that.  If intel-vaapi-driver is going to compile against X11/Xlib.h it should fail if it cannot find it on the pkg_config search.

chivakker:
To avoid breaking the build, the solution proposed by @yakuizhao makes more sense,  I'll update the pull request.

PR #51, opened 2017-02-15
Wayland-drm: Fix not finding wl_drm_interface on systems with libglvnd

jwrdegoede:
We do not want just any libEGL.so.1 we want mesa's libEGL.so.1 as that
is the only way which defines the wl_drm_interface symbol we need,
one systems with libglvnd libEGL.so.1 is a dispatcher library provided
by libglvnd and the actual mesa libEGL we want is named libEGL_mesa.so.0
so try that first.

Signed-off-by: Hans de Goede <hdegoede@redhat.com>

### DIFF
--- a/src/i965_output_wayland.c
+++ b/src/i965_output_wayland.c
@@ -36,7 +36,10 @@
 #include "i965_defines.h"
 #include "dso_utils.h"
 
-#define LIBEGL_NAME             "libEGL.so.1"
+/* We need mesa's libEGL, first try the soname of a glvnd enabled mesa build */
+#define LIBEGL_NAME             "libEGL_mesa.so.0"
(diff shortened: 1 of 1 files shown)


src/i965_output_wayland.c
```diff
@@ -380,8 +383,11 @@ i965_output_wayland_init(VADriverContextP ctx)
         goto error;
 
     i965->wl_output->libegl_handle = dso_open(LIBEGL_NAME);
-    if (!i965->wl_output->libegl_handle)
-        goto error;
+    if (!i965->wl_output->libegl_handle) {
+        i965->wl_output->libegl_handle = dso_open(LIBEGL_NAME_FALLBACK);
+        if (!i965->wl_output->libegl_handle)
+            goto error;
+    }
```
seanvk: Agree, as with the 01org/libva PR

PR #61, opened 2017-02-22
JPEG Encoder: return supported pixel formats for vaQuerySurfaceAttributes

### LINKED ISSUE #60: When vaQuerySurfaceAttributes for jpeg encoder, it doesn't return YUY2 format, which should be handled.
Since commit d540e43 landed, we can query for VAEntrypointEncPicture entrypoint. But it doesn't return YUY2 format, which jpeg encoder is capable of handling.

zzoon:
Since commit d540e43 landed, we can query for VAEntrypointEncPicture
entrypoint. But it doesn't return YUY2 format, which jpeg encoder
is capable of handling.

Fixes #60

### DIFF
--- a/src/i965_drv_video.c
+++ b/src/i965_drv_video.c
@@ -5936,6 +5936,34 @@ i965_QuerySurfaceAttributes(VADriverContextP ctx,
                   i++;
                 }
             }
+
+            /* Additional support for jpeg encoder */
+            if (obj_config->profile == VAProfileJPEGBaseline
+                    && obj_config->entrypoint == VAEntrypointEncPicture) {
+                attribs[i].type = VASurfaceAttribPixelFormat;
+                attribs[i].value.type = VAGenericValueTypeInteger;
+                attribs[i].flags = VA_SURFACE_ATTRIB_GETTABLE | VA_SURFACE_ATTRIB_SETTABLE;
+                attribs[i].value.value.i = VA_FOURCC_YUY2;
+                i++;
+
+                attribs[i].type = VASurfaceAttribPixelFormat;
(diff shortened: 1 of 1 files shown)


src/i965_drv_video.c
```diff
@@ -5935,6 +5935,12 @@ i965_QuerySurfaceAttributes(VADriverContextP ctx,
                   attribs[i].value.value.i = VA_FOURCC_I010;
                   i++;
                 }
+            } else if (obj_config->entrypoint == VAEntrypointEncPicture) {
```
scott-d-phillips: This should just be for JPEG, not all encodes.

src/i965_drv_video.c
```diff
@@ -5935,6 +5935,12 @@ i965_QuerySurfaceAttributes(VADriverContextP ctx,
...
+                attribs[i].type = VASurfaceAttribPixelFormat;
+                attribs[i].value.type = VAGenericValueTypeInteger;
+                attribs[i].flags = VA_SURFACE_ATTRIB_GETTABLE | VA_SURFACE_ATTRIB_SETTABLE;
+                attribs[i].value.value.i = VA_FOURCC_YUY2;
```
scott-d-phillips: I think we're also missing Y8, UYVY, and RGB. I think adding all these will also necessitate increasing I965_MAX_SURFACE_ATTRIBUTES.

seanvk: Yes, we need to include all the formats supported for JPEG Enc.

zzoon:
Patch updated for support YUY2, UYVY, YV16, Y800 (not RGBA as discussion on slack)

PR #77, opened 2017-03-06
configure.ac: check for AC_FUNC_ALLOCA

iamleot:
Add AC_FUNC_ALLOCA and adjust src/i965_decoder_utils.c to properly include
relevant header files for alloca(3).

Some operating systems does not have <alloca.h> (e.g. NetBSD). This fixes
compilation on such systems.

Signed-off-by: Leonardo Taccari <iamleot@gmail.com>

### DIFF
--- a/src/i965_decoder_utils.c
+++ b/src/i965_decoder_utils.c
@@ -23,7 +23,6 @@
 
 #include "sysdeps.h"
 #include <limits.h>
-#include <alloca.h>
 
 #include "intel_batchbuffer.h"
 #include "intel_media.h"
@@ -343,7 +342,7 @@ avc_get_first_mb_bit_offset_with_epb(
     if (buf_size > data_size)
         buf_size = data_size;
 
-    buf = alloca(buf_size);
+    buf = malloc(buf_size);
     ret = dri_bo_get_subdata(
         slice_data_bo, slice_param->slice_data_offset,
         buf_size, buf
@@ -354,6 +353,8 @@ avc_get_first_mb_bit_offset_with_epb(
         if (buf[i] == 0x03 && buf[i - 1] == 0x00 && buf[i - 2] == 0x00)
             i += 2, j++, n++;
     }
+
+    free(buf);
 
     out_slice_data_bit_offset = in_slice_data_bit_offset + n * 8;
 
--- a/configure.ac
+++ b/configure.ac
@@ -1,7 +1,7 @@
 # intel-vaapi-driver package version number
 m4_define([intel_vaapi_driver_major_version], [1])
 m4_define([intel_vaapi_driver_minor_version], [8])
-m4_define([intel_vaapi_driver_micro_version], [0])
+m4_define([intel_vaapi_driver_micro_version], [1])
 m4_define([intel_vaapi_driver_pre_version],   [1])
 m4_define([intel_vaapi_driver_version],
           [intel_vaapi_driver_major_version.intel_vaapi_driver_minor_version.intel_vaapi_driver_micro_version])


yakuizhao:
It seems that the alloca is used for the buffer allocation that is automatically freed.
Maybe we can remove the alloca and manage the required buffer manually.
Will you please use the malloc/free instead of alloca and then remove the header file of "alloca.h"?

src/i965_decoder_utils.c
```diff
@@ -23,7 +23,17 @@
 
 #include "sysdeps.h"
 #include <limits.h>
-#include <alloca.h>
+#ifdef STDC_HEADERS
+# include <stdlib.h>
```
xhaihao: sysdeps.h has already included stdlib.h, it would be better to move this check to sysdeps.h

xhaihao:
@yakuizhao for some cases, alloca is more efficient than malloc/free

yakuizhao:
@xhaihao , sometimes the alloca seems higher efficiency than malloc/free.  But based on the user-manual of alloca, it is machine and compiler-dependent.  Except the longjmp/siglongjmp, the alloca is discouraged.
So I suggest that it can be replaced.

xhaihao:
Okay, It is fine for me to replace alloca with malloc/free. the code is only used for some legacy platforms.

iamleot:
Hello Xiang,

"Xiang, Haihao" writes:
> xhaihao commented on this pull request.
> 
> 
> 
> > @@ -23,7 +23,17 @@
>  
>  #include "sysdeps.h"
>  #include <limits.h>
> -#include <alloca.h>
> +#ifdef STDC_HEADERS
> +# include <stdlib.h>
> 
> sysdeps.h has already included stdlib.h, it would be better to move this check to sysdeps.h
> 

Sorry, I have not noticed it. I will address that (but please see
further comments)!


Thanks for pointing that out!

iamleot:
Hello Xiang,

"Xiang, Haihao" writes:
> Okay, It is fine for me to replace alloca with malloc/free. the code is only used for some legacy platforms.

Allright, I will try to replace alloca(3) with malloc(3)/free(3) as
discussed in the next hours (probably not before 20:00 UTC or so).


Thank you very much for yours and yakuizhao's suggestions about it!

iamleot:
Hello @xhaihao and @yakuizhao,
I have just tried to address it as discussed and tested on NetBSD.

Please let me know if any further adjustments are needed!


Thank you very much again!

PS: Not checking malloc(3) return value should be correct in that case
because that's already handled in dri_bo_get_subdata().

xhaihao:
@iamleot could you squash the two commits into one commit for this issue?

iamleot:
Hello Xiang,

"Xiang, Haihao" writes:
> @iamleot could you squash the two commits into one commit for this issue?
> 

Sure, I have just done it! (that's the first time that I've used
`git rebase -i`, so please let me know if I have m[ei]ssed something!)

src/i965_decoder_utils.c
```diff
@@ -355,6 +354,9 @@ avc_get_first_mb_bit_offset_with_epb(
             i += 2, j++, n++;
     }
 
+    if (buf)
+        free(buf);
+
```
xhaihao: free() can accept NULL pointer, so it is unnecessary to check the pointer against NULL.

iamleot:
Hello Xiang,

"Xiang, Haihao" writes:
> > @@ -355,6 +354,9 @@ avc_get_first_mb_bit_offset_with_epb(
>              i += 2, j++, n++;
>      }
>  
> +    if (buf)
> +        free(buf);
> +
> 
> free() can accept NULL pointer, so it is unnecessary to check the pointer against NULL.
> 

OK! I have just adjusted it and squashed the various commits again.

xhaihao:
@iamleot Could you rebase your branch against the latest master branch in https://github.com/01org/intel-vaapi-driver.git ? I got conflict when doing rebase.  In addition, I still see a few new commits besides the merge commits  in your branch, could you squash your fixes into a single commit?

iamleot:
Hello Xiang,

"Xiang, Haihao" writes:
> @iamleot Could you rebase your branch against the latest master branch in https://github.com/01org/intel-vaapi-driver.git ? I got conflict when doing rebase.  In addition, I still see a few new commits besides the merge commits  in your branch, could you squash your fixes into a single commit?

OK! I have just rebased to the latest master branch and squashed the
commits again.

Issue #82, opened 2017-03-10
Fix possible garbage data after malloc in i965_create_buffer_internal function


rislim:
1. Fix memset possible happen incorrect size value, it should be using msize instead of size. After align, the size value may increase if size value is not aligned with 4.
2. Avoid possible garbage data, recommend do a memset after malloc to initialize the value to zero.

PR #83, opened 2017-03-10
Fix some possible garbage data after malloc in i965_create_buffer_int…

rislim:
…ernal func.

Fixes https://github.com/01org/intel-vaapi-driver/issues/82

Signed-off-by: Lim Siew Hoon <siew.hoon.lim@intel.com>

### DIFF
--- a/src/i965_drv_video.c
+++ b/src/i965_drv_video.c
@@ -2532,11 +2532,11 @@ i965_create_buffer_internal(VADriverContextP ctx,
         assert(buffer_store->buffer);
 
         if (!wrapper_flag) {
+            memset(buffer_store->buffer, 0, msize * num_elements);
             if (data)
                 memcpy(buffer_store->buffer, data, size * num_elements);
-            else
-                memset(buffer_store->buffer, 0, size * num_elements);
-        }
+        } else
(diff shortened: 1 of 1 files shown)


src/i965_drv_video.c
```diff
@@ -2532,11 +2532,11 @@ i965_create_buffer_internal(VADriverContextP ctx,
...
 
         if (!wrapper_flag) {
+            memset(buffer_store->buffer, 0, msize * num_elements);
             if (data)
                 memcpy(buffer_store->buffer, data, size * num_elements);
-            else
-                memset(buffer_store->buffer, 0, size * num_elements);
-        }
+        } else
+            memset(buffer_store->buffer, 0, 4);
```
yakuizhao: The allocated four bytes are meaningless when the buffer is allocated with wrapper flag. 
So it doesn't matter that it is not initialized to zero.

At the same time I don't think that it is necessary to initialize the buffer again when it needs to copy the passed data.
When the data_pointer is passed, the data length is consistent with the total size of allocated buffer.  
So I think that this patch is not needed.

xhaihao:
#82 is closed as invalid

PR #86, opened 2017-03-13
configure: change package tarball name to use hyphens

### LINKED ISSUE #85: pre-release and release tarballs should comply with naming convention using hyphens
All our previous releases use hyphens for the final tarball name.  To reduce the amount of changes on the distributions building scripts the names should be using hyphens.

The naming can be checked on the old repository https://freedesktop.org/software/vaapi/testing/libva-intel-driver/

chivakker:
Compilation scripts on different distributions (i.e. arch, gentoo) rely
on the tarball name to use hyphens between words.

This patch will create tarballs with hyphenated names and keep the
legacy package naming

TEST="make dist, should produce intel-vaapi-driver.tar.*"

Fixes #85

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -20,7 +20,7 @@ AC_PREREQ([2.57])
 AC_INIT([intel_vaapi_driver],
         [intel_vaapi_driver_version],
         [https://github.com/01org/intel-vaapi-driver/issues/new],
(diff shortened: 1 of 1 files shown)


yakuizhao:
make sense and LGTM

PR #91, opened 2017-03-15
Check pointers against NULL before using

xhaihao:
This fixes https://github.com/01org/intel-vaapi-driver/issues/90


yakuizhao:
This patch looks good to me.
And it will be better that more checks are added for the following buffer:
    >i965_gpe_map_curbe() in i965_encoder_vp8.c/gen9_avc_encoder.c
    >i965_gpe_map_resource() in i965_encoder_vp8.c/gen9_avc_encoder.c

Thanks

xhaihao:
Thanks, I added more checks.

PR #94, opened 2017-03-20
HEVC encoding quality improvement on SKL/APL/KBL

### LINKED ISSUE #95: HEVC encoding quality improvement on SKL/APL/KBL
Here re-arch the hevc encoder on SKL/APL/KBL and achieve better quality than legacy hevc encoder

pengche1:
Fixes #95


src/gen9_hevc_enc_kernels.h
```diff
@@ -0,0 +1,2318 @@
...
+    GEN9_ENC_WP,
+    GEN9_ENC_SFD,
+    GEN9_ENC_MBENC_I_LUMA,
+    GEN9_ENC_MPU,
+    GEN9_ENC_TPU,
+    GEN9_ENC_SCALING_CONVERSION,
+    GEN9_ENC_DYS
+} GEN9_ENC_OPERATION;
+
+#define GEN9_HEVC_ENC_KERNEL_SIZE 147456
```
yakuizhao: Please split the kernel binary and the structure definition into two files instead of combining them together.

pengche1: I should have moved kernel binary to gen9_hevc_enc_kernels.c. So do I have some misunderstanding?

src/gen9_hevc_encoder.c
```diff
@@ -0,0 +1,7361 @@
...
+                                 struct gpe_media_object_walker_parameter *gpe_param,
+                                 int split_count,
+                                 unsigned int max_slight_height,
+                                 int use_hw_scoreboard,
+                                 int scoreboard_type)
+{
+    int width = priv_state->width_in_mb;
+    int height = max_slight_height * 2;
+    int ts_width = ((width + 3) & 0xfffc) >> 1;
+    int lcu_width = (width + 1) >> 1
```
yakuizhao: Now the file of gen6_mfc_common.c has one similar definition of map_44_lut_value.
    >intel_format_lutvalue
Can we reuse it ?

yakuizhao: please avoid the static definition in the function.

yakuizhao: Is it necessary to initialize the buffer of res_mb_code_surface?
Please avoid the unnecessary initialization.

yakuizhao: Please double check the above command.

yakuizhao: Please configure the correct cache type for the buffer.

yakuizhao: Please don't always use the flat scaling list. It should be based on the application configuration.

yakuizhao: Please assure that res_pak_slice_batch_buffer is not accessed by CPU any more. In such case this helps to assure that the buffer domain is configured correctly.

src/gen9_hevc_enc_kernels.h
yakuizhao: the  structure definition should be defined in one dedicated file while the kernel binary is defined in another file.

yakuizhao: the  structure definition should be defined in one dedicated file while the kernel binary is defined in another file.

src/gen9_hevc_enc_const_def.h
```diff
@@ -0,0 +1,528 @@
...
+ *    Chen, Peng <chen.c.peng@intel.com>
+ *
+ */
+
+#ifndef GEN9_HEVC_ENCODER_CONST_DEF_H
+#define GEN9_HEVC_ENCODER_CONST_DEF_H
+
+#define GEN9_HEVC_ENC_MEMETHOD_TABLE 0
+
+static const unsigned int GEN9_HEVC_ME_METHOD[NUM_TU_MODES] =
```
pengche1: I have different idea. If it is not prefixed with static, this variable will be a global variable. It would pollute the whole namespace and it would have some issues if other project also use this variable name as golobal variable. by the way, it should be only used in one file, no duplicated definition is needed here

src/gen9_hevc_enc_utils.h
```diff
@@ -0,0 +1,47 @@
...
+#include <i915_drm.h>
+#include <intel_bufmgr.h>
+
+#include <va/va.h>
+
+static const char hevc_qpc_table[22] = {
+    29, 30, 31, 32, 32, 33, 34, 34, 35, 35, 36, 36, 37, 37, 37, 38, 38, 38, 39, 39, 39, 39
+};
+
+unsigned int
```
pengche1: ok, it will be fixed. Thanks!

src/gen9_hevc_encoder.c
```diff
@@ -0,0 +1,7361 @@
...
+
+#define ALLOC_GPE_RESOURCE(RES, NAME, SIZE)                 \
+    do{                                                     \
+        i965_free_gpe_resource(&priv_ctx->RES);             \
+        if (!i965_allocate_gpe_resource(i965->intel.bufmgr, \
+                                 &priv_ctx->RES,            \
+                                 SIZE,                      \
+                                 NAME))                     \
+            goto FAIL;                                      \
+        i965_zero_gpe_resource(&priv_ctx->RES);             \
```
pengche1: it is only called once in the initialization, not for each frame encoding. so it impacts the performance really a little

src/gen9_hevc_encoder.c
```diff
@@ -0,0 +1,7361 @@
...
+    struct gen9_hevc_encoder_state *priv_state = NULL;
+    int res_size = 0, size_shift = 0;
+    int width = 0, height = 0;
+
+    vme_context = (struct encoder_vme_mfc_context *)encoder_context->vme_context;
+    priv_ctx = (struct gen9_hevc_encoder_context *)vme_context->private_enc_ctx;
+    generic_state = (struct generic_enc_codec_state *)vme_context->generic_enc_state;
+    priv_state = (struct gen9_hevc_encoder_state *)vme_context->private_enc_state;
+
+    if (priv_ctx->res_inited)
```
pengche1: Ok, it will be enhanced. Thanks!

src/gen9_hevc_encoder.c
```diff
@@ -0,0 +1,7361 @@
...
+    if (priv_ctx->scaled_2x_surface_obj)
+          i965_DestroySurfaces(priv_ctx->ctx, &priv_ctx->scaled_2x_surface_id, 1);
+
+    width = priv_state->frame_width_in_max_lcu >> 1;
+    height = priv_state->frame_height_in_max_lcu >> 1;
+    if (priv_state->bit_depth_luma_minus8) {
+        width = ALIGN(width, 32);
+        height = ALIGN(height, 32);
+    }
+
```
pengche1: No. 2XScaling is a new process in HEVC MBEnc. This surface is used by later MBEnc process

src/gen9_hevc_enc_const_def.h
```diff
@@ -0,0 +1,528 @@
...
+        0x0000, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000, 0x000c,
+        0x000c, 0x0020, 0x0020, 0x0033, 0x0033, 0x0045, 0x0045, 0x0057,
+        0x0057, 0x006b, 0x006b, 0x0081, 0x0081, 0x009a, 0x009a, 0x00b8,
+        0x00b8, 0x00db, 0x00db, 0x0104, 0x0104, 0x0135, 0x0135, 0x016d,
+        0x016d, 0x01af, 0x01af, 0x01fb, 0x01fb, 0x0252, 0x0252, 0x02b6,
+        0x02b6, 0x0326, 0x0326, 0x03a5, 0x03a5, 0x0432, 0x0432, 0x04d0,
+        0x04d0, 0x057e, 0x057e, 0x062c
+    }
+};
+
```
yakuizhao: Although some tables/definitions are only used for HEVC encoding on SKL/BXT, it is possible that some other tables/definitions can be used for the encoding on other platform. 
If they are included in two or more files, multiple definitions are added.

src/gen9_hevc_encoder.c
yakuizhao: Please configure the correct cache type for the buffer

yakuizhao: Please don't use the flat scaling list under all the scenarios. It should be configured based on middleware.

yakuizhao: please follow the passed parameter to configure the IRF_IDX_STATE correctly. For example: weight flag.

yakuizhao: Please assure that  the buffer domain of res_pak_slice_batch_buffer is configured correctly when the command buffer is submitted.

yakuizhao: please check whether the buffer domain is configured

yakuizhao: please double check the above command.

src/gen9_hevc_enc_utils.c
```diff
@@ -0,0 +1,122 @@
...
+ * next paragraph) shall be included in all copies or substantial portions
+ * of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+ * OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT.
+ * IN NO EVENT SHALL PRECISION INSIGHT AND/OR ITS SUPPLIERS BE LIABLE FOR
+ * ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+ * TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWAR
```
QuPengfei: license header is incomplete.

QuPengfei: license header is incomplete.

src/gen9_hevc_enc_utils.h
```diff
@@ -0,0 +1,47 @@
...
+ * next paragraph) shall be included in all copies or substantial portions
+ * of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+ * OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT.
+ * IN NO EVENT SHALL PRECISION INSIGHT AND/OR ITS SUPPLIERS BE LIABLE FOR
+ * ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+ * TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWAR
```
QuPengfei: license header is incomplete

src/gen9_hevc_encapi.h
```diff
@@ -0,0 +1,42 @@
...
+
+#define HEVC_ENC_BETTER_QUALITY         1
+
+extern Bool
+gen9_hevc_vme_context_init(VADriverContextP ctx, struct intel_encoder_context *encoder_context);
+
+extern Bool
+gen9_hevc_pak_context_init(VADriverContextP ctx, struct intel_encoder_context *encoder_context);
+
+#endif  // _INTEL_GEN9_HEVC_ENC_API_H_
```
QuPengfei: it is better put these two functions in the file i965_encoder_api.h and remove this file

src/i965_encoder.c
```diff
@@ -808,8 +810,28 @@ intel_encoder_check_misc_parameter(VADriverContextP ctx,
         VAEncMiscParameterBufferQualityLevel* param_quality_level = (VAEncMiscParameterBufferQualityLevel*)pMiscParam->data;
         encoder_context->quality_level = param_quality_level->quality_level;
 
-        if (encoder_context->quality_level == 0)
-            encoder_context->quality_level = ENCODER_DEFAULT_QUALITY;
+        if (encoder_context->quality_level == 0) {
```
QuPengfei: default quality level value in the function intel_enc_hw_context_init() is set to 4. so here if(==0) will never be called. how about removing these changes?

src/gen9_hevc_enc_utils.c
pengche1: ok, it will be fixed. thanks

src/gen9_hevc_encapi.h
pengche1: ok, i will move the hevc apis to this common file

src/i965_encoder.c
pengche1: no, this quality_level is from extern APIs and can change the value set by intel_enc_hw_context_init()  . And value 0 means driver should choose its own quality level automatically

pengche1:
hi all, I have pushed a new version for review. any comments, just tell me. thanks!

src/gen9_hevc_enc_utils.h
```diff
@@ -0,0 +1,47 @@
...
+#ifndef GEN9_HEVC_ENCODER_UTILS_H
+#define GEN9_HEVC_ENCODER_UTILS_H
+
+#include <drm.h>
+#include <i915_drm.h>
+#include <intel_bufmgr.h>
+
+#include <va/va.h>
+
+static const char hevc_qpc_table[22] = {
```
yakuizhao: For the static table, I think that it can be declared/defined in *.h/*c.
Otherwise when this file is included twice, there exists the duplicated definition.

src/gen9_hevc_encoder.c
```diff
@@ -0,0 +1,7376 @@
...
+            if (!surface_priv->surface_obj_nv12)
+                return VA_STATUS_ERROR_ALLOCATION_FAILED;
+
+            i965_check_alloc_surface_bo(ctx, surface_priv->surface_obj_nv12, 1,
+                                        VA_FOURCC('N', 'V', '1', '2'), SUBSAMPLE_YUV420);
+        }
+
+        surface_priv->surface_reff = obj_surface;
+
```
yakuizhao: Please move the below definition earlier to avoid the memory leak when it fails on buffer allocation.

src/gen9_hevc_encoder.c
```diff
@@ -0,0 +1,7376 @@
...
+    i965_DestroySurfaces(surface_priv->ctx, &surface_priv->surface_id_nv12, 1);
+    surface_priv->surface_id_nv12 = VA_INVALID_SURFACE;
+    surface_priv->surface_obj_nv12 = NULL;
+
+    dri_bo_unreference(surface_priv->motion_vector_temporal_bo);
+
+    free(surface_priv);
+}
+
```
yakuizhao: The current code doesn't consider the change of width/height for the encoded frame especially when the encoded frame is also sued when switching from higher resolution to lower resolution.

src/gen9_hevc_encoder.c
yakuizhao: Although it is initialized only once, the allocated buffer/surface is quite large. And most buffers are only accessed by GPU,  I don't think that it is necessary to initialize all of them.

src/gen9_hevc_encoder.c
```diff
@@ -0,0 +1,7376 @@
...
+        gpe_param->local_end.y = 0;
+    }
+}
+
+static void
+gen9_hevc_init_object_walker_26z(struct gen9_hevc_encoder_state *priv_state,
+                                 struct i965_gpe_context *gpe_context,
+                                 struct gpe_media_object_walker_parameter *gpe_param,
+                                 int split_count,
+
```
yakuizhao: It will be better to use the explicit definition for BTI definition. The usage of bti_idx++ is difficult to understand the meaning.

yakuizhao: How about defining it out of function?

pengche1: ok, thanks!

pengche1: It is from HLD reference doc. It is more safe to keep it

src/gen9_hevc_encoder.c
pengche1: It has already implemented in the driver. please check priv_ctx->res_inited, it will be cleared for new picture width or height from SPS

src/gen9_hevc_encoder.c
pengche1: ok, thanks!

src/gen9_hevc_enc_utils.h
pengche1: ok, i will move this table into file gen9_hevc_encoder.c

pengche1:
Please help me to review this new RC4. Any comments, just let me know. Thanks!

YuJiankang:
LGTM

xhaihao:
Could you squash the last commit into other commits? otherwise multi slice encoding will cause critical issue with the commits before the last commit.

pengche1:
sure, thanks

PR #96, opened 2017-03-20
AVC encoder rework on kbl  

QuPengfei:
This series of patch improve the AVC encoder on KBL


QuPengfei:
patch still looks out of order. :(

QuPengfei:
change the commit date and update.

yakuizhao:
It seems that the avc encoding on SKL causes the B-frame issues.
It will be better that the issue is fixed firstly.

QuPengfei:
@yakuizhao Did you see the B frames issue on SKL? or you mean sree mentioned about gstreamer-vaapi?
I do not think it is driver issue. our internal test did not show this issue.

Issue #84, opened 2017-03-10
avcenc produces run-to-run bit compare mismatches.


tmjohns1:
Sporadically seeing sha1sum mismatches from run-to-run of avcenc (see below).  Reproduced with Big Buck Bunny clip, as well as 176x144-25_1882_seq4.yuv.  Pengfei has reproduced this issue.

Typical invocation:
avcenc 1920 1080 big_buck_bunny_1080p24_frames_11858_12100.yuv bbb_1080p24-cqp10-1.264 --qp=10 --mode=2

46668c54572db0e02024ed8f1663f892d4cbe966  bbb_1080p24-cqp10-1.264
0fe29c90b887403e52c66fc7d834b3802c82182e  bbb_1080p24-cqp10-2.264
ef6b2e0ca63d1ca6f4e7d309671658e0b48d4c03  bbb_1080p24-cqp10-3.264
4758d7207c7365a579ca524ba60332b8e05d5386  bbb_1080p24-cqp10-4.264
cd83106cd2253e7897358ef2c495b9a6f80a57a5  bbb_1080p24-cqp10-5.264

ed064d19d758d23c64ac30bd6a20e4f541658adb  bbb_1080p24-cqp20-1.264
ed064d19d758d23c64ac30bd6a20e4f541658adb  bbb_1080p24-cqp20-2.264
ed064d19d758d23c64ac30bd6a20e4f541658adb  bbb_1080p24-cqp20-3.264
ed064d19d758d23c64ac30bd6a20e4f541658adb  bbb_1080p24-cqp20-4.264
ed064d19d758d23c64ac30bd6a20e4f541658adb  bbb_1080p24-cqp20-5.264

c797c1b0bceee677ecdb6be740c7c6984f56414d  bbb_1080p24-cqp25-1.264
c797c1b0bceee677ecdb6be740c7c6984f56414d  bbb_1080p24-cqp25-2.264
c797c1b0bceee677ecdb6be740c7c6984f56414d  bbb_1080p24-cqp25-3.264
c797c1b0bceee677ecdb6be740c7c6984f56414d  bbb_1080p24-cqp25-4.264
c797c1b0bceee677ecdb6be740c7c6984f56414d  bbb_1080p24-cqp25-5.264

d65807d292bbf4d1e29d8b3a828183515e276099  bbb_1080p24-cqp30-1.264
d65807d292bbf4d1e29d8b3a828183515e276099  bbb_1080p24-cqp30-2.264
[...]

PR #102, opened 2017-03-24
this fix the B frame issue for AVC high profile

QuPengfei:


fhvwy:
You can't rewrite the PPS here.  Containers with global headers (e.g. MP4 and all similar ISO BMFF-derived formats) and anything which transmits the parameters sets out-of-band (e.g. SDP for RTP) will already have written/sent the existing PPS and will break if the stream itself then uses different parameters.

sreerenjb:
What Mark saying is right. We are not supposed to generate the SPS/PPS from driver. The whole AVCDecoderConfigurationRecord (which includes sps, pps and vui) could be already generated in upper layer even before the actual encoding starts, and it is perfectly valid as per Specifications too.

Note: In case if you are testing with gstreamer, you might not see this issue. Becuase gstreamer test cases that we use always parse the encoded data(after the encoding) and generate the sps/pps based on it.
But don't assume that it is the general rule or usage scenario always.

xhaihao:
Agreed with @fhvwy @sreerenjb, we should fix or workaround it in the driver in another way.

QuPengfei:
@fhvwy @sreerenjb @xhaihao agreed with you. thanks to point out this. i will give a simple WA and not generate the PPS bitstrem in the driver.

sreerenjb:
No, we can't fix this issue by just disabling it in driver. Because the middleware has no clue about this internal disabling and it generate the pps header with 8x8 flag as TRUE which simply make the stream undecodable.

yakuizhao:
@sreerenjb, The 8x8 flag is defined as two levels. Transform_mode_flag in picture_parameter only provides the hint that each macroblock has the possiblity of using 8x8 for residual transform.  Each macroblock will determine whether 8x8 or 4x4, which is affected by the transform_mode_flag in picture_parameter. That is to say:  when the transmode_flag_flag in picture_parameter is 8x8,  it is still ok that 4x4 is used for each macroblock(8x8 is optional).  
In such case the upper middleware can still generate the PPS header with 8x8 flag enabled.  The WA patch in driver only  control the selection of 8x8 and 4x4 for each macroblock.

QuPengfei:
Agree with @yakuizhao.

@sreerenjb,  for VME, the driver configure with transform_8x8_mode_flag=0 and still keep original value(1) for Image state in the PAK. so there is no issue for decoding.

BTW, transform_8x8_mode_flag means to use 8x8 as hint, but it is not necessary.

xhaihao:
@QuPengfei @yakuizhao are right, even if transform_8x8_mode_flag is 1, we can set transform_size_8x8_flag to 0 in macroblock level

sreerenjb:
I am Okay with the work around as long as the PAK still using the transform_8x8 as TRUE and the decoder can still use the correct transform coefficient decoding process using section 8.5.

Just noticed that the B-frame encoding artifacts are still present, but this could be because of #84 .
 I can see visible artifacts when enabling B-frame encoding(irrespecive of 8x8 transform) which simply disappear If I revert back to commit 1d43c90ae9f0cf40cabca11bd9f5abb06a31d8efx`

sreerenjb:
This artifact issue is due to a bug in gstreamer-vaapi which I am using for testing. It was working with old media kernel because that code path was not accessing the POC directly from VAEncSliceParameterBufferH264->RefPicList{0,1}. But the new code is directly accessing this poc value for scale factor calculation and unfortunately, gstreamer-vaapi is populating the POC value only in VAEncSliceParameterBufferH264, not in RefPicLists.

I will fix this in gstreamer-vaapi.

fhvwy:
This change is a large regression in the coding performance of high-resolution sequences.  In my testing in constant quality mode with B-frames disabled a typical 4K stream might increase in size by ten percent or more.  (Higher resolutions are affected to a greater degree because individual macroblocks tend to contain less detail; at 1080p it is a few percent at most.)

So, can someone explain what the underlying issue actually is here - is this a hardware bug affecting all gen9 cores, or is it a software problem with the shader blobs which can be fixed later?

In mitigating the problem right now, would it be possible to restrict this change to avoid the 8x8 transform only when coding B-frames (coding I/P-frames in the same stream normally), or if that isn't possible only when B-frames are present in the stream?

chivakker:
@fhvwy this is a bug that we're looking into solving pretty soon and it is not a h/w bug.  We will revert this work around and provide a fix as soon as possible. 

We can certainly look into mitigate the problem by avoid dct8x8 only with B frames.

sreerenjb:
@fhvwy I guess you are testing with "brames=0 dct8x8=1 profile=high ratecontrol=cbr" right? I haven't seen something like 10% size increase when testing with/without the proposed patch. 

But interestingly enabling B frames producing a video of higher size than a stream without B frames :)

sreerenjb:
Anyway, Here is the gstreamer fix: https://bugzilla.gnome.org/show_bug.cgi?id=780620
Hopefully, I will push the fix tomorrow.

yakuizhao:
@fhvwy , What Chivakker said is right. It is not HW bug. It is only the software problem when configuring the parameter for media kernel.  Sure that this will be fixed ASAP.

QuPengfei:
@fhvwy . yes, we can try the method to set transform_8x8_flag=0 only for high profile B frame. Ideally, it should be helpful for files increment issue when 4K.

BTW, this patch does not affect the main profile.

fhvwy:
@sreerenjb  Yes, pretty much.  (Assuming you actually mean CQP - CBR obviously would give essentially the same sizes.)

E.g. with avconv, "-c:v h264_vaapi -bf 0 -qp 30 -profile <whatever> -frames:v 5000" and a 4K input http://distribution.bbb3d.renderfarming.net/video/mp4/bbb_sunflower_2160p_60fps_normal.mp4 :

```
Size        Profile  Change
106019894   High     Before
119708392   High     After
119598301   Main     Before
119598301   Main     After
```

Main profile is unchanged as expected, but high has regressed to be worse than main (presumably because the coding tools are now equivalent, and there is a bit of extra overhead for always choosing against using the 8x8 transform in the macroblock layer).

sreerenjb:
Typo, sorry :), CQP of course...

PR #103, opened 2017-03-24
Fix: wrong type define and cause the memory access error

### LINKED ISSUE #101: wrong type define and cause the memory access error

QuPengfei:
Fixes #101


xhaihao:
LGTM

PR #120, opened 2017-04-05
ENC: Disable the ROI feature of AVC encoder on SKL/APL/KBL

### LINKED ISSUE #108: ROI(Region of Interest) of avc encding is broken on the master on SKL/APL/KBL
Since the commit e9a0038 landed, it seems that ROI is broken.
In addition, a size of produced h264 data with roi is much more (over ten times) than h264 data's size without roi. Is this expected?

I tested with libva-utils/encode/avcenc and KBL.

QuPengfei:
this is WA to disable the feature as the new improved AVC encoder does not support it by now.

Fixes #108


src/i965_drv_video.c
```diff
@@ -1043,18 +1043,23 @@ i965_GetConfigAttributes(VADriverContextP ctx,
                     profile == VAProfileH264Main ||
                     profile == VAProfileH264High) {
 
-                    VAConfigAttribValEncROI *roi_config =
-                        (VAConfigAttribValEncROI *)&(attrib_list[i].value);
+                    if (IS_GEN9(i965->intel.device_info))
```
scott-d-phillips: I don't believe this should disable ROI for EncSliceLP

yakuizhao: Agree with what Scott said.
Please only disable it for non-supported platforms/

PR #121, opened 2017-04-05
ENC: disable MVC feature of AVC encoder on SKL/APL/KBL

### LINKED ISSUE #115: MVC feature of AVC encoder is broken on SKL/APL/KBL

QuPengfei:
this is WA to disable the MVC feature as the new improved AVC encoder does not support it by now.

Fixes #115

we will add back this feature in Q2 release


src/i965_drv_video.c
```diff
@@ -618,7 +618,8 @@ i965_QueryConfigEntrypoints(VADriverContextP ctx,
        if (HAS_H264_MVC_DECODING_PROFILE(i965, profile))
             entrypoint_list[n++] = VAEntrypointVLD;
 
-       if (HAS_H264_MVC_ENCODING(i965))
+       if (HAS_H264_MVC_ENCODING(i965) &&
```
yakuizhao: This check looks too hack.
The MVC encoding can be disabled in skl/kbl_hw_codec_info.

QuPengfei: good comments. thanks to point out.  I will update the patch.

PR #122, opened 2017-04-05
ENC: add AUD nal unit at the beginning of picture

### LINKED ISSUE #110: VAAPI driver producing non-conformant AVC bitstream
When using VAAPI driver with Media SDK, it seems to re-order the AU delimiter following the Supplemental Enhancement Information.  I believe this is in violation of the spec:

“The following constraints shall be obeyed by the order of the coded pictures and non-VCL NAL units within an access unit:
– When an access unit delimiter NAL unit is present, it shall be the first NAL unit. There shall be at most one access unit delimiter NAL unit in any access unit”.

Typical example of conformant stream:
00 00 01 _09 10_ 00 00 00 01 27 64 00 0d ac 2b 60
a8 47 c9 58 08 80 00 00 03 00 80 00 00 19 4e d0
e1 57 00 00 00 01 28 ee 3c b0 00 00 00 01 06 00
01 c0 01 01 04 80 00 00 01 25 88 80 41 1f cd

Typical example of non-conformant stream:
00 00 00 01 27 64 00 0d ac 2b 60 a8 47 c9 58 08
80 00 00 03 00 80 00 00 19 4e d0 e1 57 00 00 00
01 28 ee 3c b0 00 00 00 01 06 00 01 c0 01 01 04
80 00 00 00 01 _09 10_ 00 00 01 25 88 80 41 1f cd

It was confirmed by code inspection that MSDK forms the headers and sends them to the driver in following order: AU delimiter, SPS, PPS, SEI, slice headers.  The OTC driver seems to internally change the order.

Attempts to create a avcenc or yamitranscode command line to generate AUs have been unsuccessful.

QuPengfei:
Fixes #110


sreerenjb:
The patch Okay. Only problem is in the middleware, in terms of backward compatibility :)
Please check the below comment:
https://bugzilla.gnome.org/show_bug.cgi?id=776712#c6

But from driver perspective, we should consider it as a bug fix. Becuase as I mentioned in the above link, AUD will get inserted even without this driver patch, but *only after SPS/PPS/SEI* , which is against what h264 spec recommend.  May be good to change the commit message saying "fix the order of AUD insertion"

The whole issue is because of the way we handle PackedHeaderRawData. As per VA spec "Packed raw data header can be used by the client to insert a header into the bitstream data buffer at the point it is passed". Ideally, we should respect the order in which the middleware submitting the packed headers, instead of hard coding the order like "aud, sps, pps, rawdata_hdr0_slice0, rawdata_hdr1_slice0,....raw_data_slice0,"

src/gen9_avc_encoder.c
```diff
@@ -6648,6 +6648,51 @@ gen9_mfc_avc_insert_object(VADriverContextP ctx,
...
+    for (i = 0; i < count; i++) {
+        unsigned int skip_emul_byte_cnt;
+
+        header_data = (unsigned int *)encode_state->packed_header_data_ext[start_index + i]->buffer;
+        nal_type = (unsigned char *)header_data;
+
+        param = (VAEncPackedHeaderParameterBuffer *)(encode_state->packed_header_params_ext[start_index + i]->buffer);
+
+        /* skip the slice header packed data type as it is lastly inserted */
+        if (param->type != VAEncPackedHeaderRawData && *(nal_type+4) != AVC_NAL_DELIMITER )
```
yakuizhao: The check of nal_unit_type is too hack.
It will be better to parse it by using generic style.

BTW:  It is enough to check the nal_unit_type.

xhaihao: Agreed with @yakuizhao , 
1. The start code prefix is 000001, not 00000001, so (nal_type+4) may not be the byte used for nal_unit_type.
2. It is legal to have some leading zero before the start code prefix, 000000000109
3. only the low 5bits is used as nal_unit_type

src/gen9_avc_encoder.c
```diff
@@ -6672,11 +6718,12 @@ gen9_mfc_avc_insert_slice_packed_data(VADriverContextP ctx,
         unsigned int skip_emul_byte_cnt;
 
         header_data = (unsigned int *)encode_state->packed_header_data_ext[start_index + i]->buffer;
+        nal_type = (unsigned char *)header_data;
 
         param = (VAEncPackedHeaderParameterBuffer *)(encode_state->packed_header_params_ext[start_index + i]->buffer);
 
         /* skip the slice header packed data type as it is lastly inserted */
-        if (param->type == VAEncPackedHeaderSlice)
+        if (param->type == VAEncPackedHeaderSlice || *(nal_type+4) == AVC_NAL_DELIMITER)
```
xhaihao: Need a similar change mentioned above to the check for nal unit type.

fhvwy:
Why is this being hacked in here?  The bitstream metadata elements like AUD and SEI should not be the responsibility of the hardware driver at all.  If they are desired by some particular use-case, then the user can trivially insert them in the correct place themselves (it's easy to add things to the beginning of every access unit, and they know all of the slice types already).

As an alternative, some sort "start of access unit" and "end of access unit" packed headers could be added to VAAPI, which the user could then set appropriately to insert whatever set of the delimiting NAL units that they need for their use-case (access-unit-delimiter, end-of-sequence, end-of-stream).

sreerenjb:
@fhvwy the above patch is not for "create & insert AUD" from driver, but it is for "inserting the application provided AUD" to the stream. 

As I mentioned above, even with out this patch AUD will get inserted, but would be in a wrong place. 
So the patch is a *bug fix to insert AUD in correct position*

sreerenjb:
@fhvwy The ideal solution could be to rewrite the entire packed header management in order to respect the order in which application submit RawData header. But that would probably too much rewrite. :)

fhvwy:
@sreerenjb  Oops, sorry, I misread that completely.  (Can't say I ever thought of submitting multiple RawData headers.)

I'm not sure the user submission ordering is necessarily what you want - the position in the access unit /is/ precisely defined by the NAL unit type (for H.265, too, where suffix SEI is also present to make it more inconvenient).  In any case, if this is intended to be a common feature then it should probably work beyond gen9 and the libva headers want to be updated to reflect that RawData works for non-SEI NAL units.

sreerenjb:
@fhvwy Rawdata headers are not just for the non vcl headers. WE need RawData headers for profiles like MVC and SVC where we can simply stuff specific nal units (even slices) where ever the application want. Even the VA spec saying it "Packed raw data header can be used by the client to insert a header into the bitstream data buffer at the point it is passed". So I would say, even if specification has a tight constraint in the ordering of NALs, We should respect the application submission order instead of hard coding the order inside driver. Also we have to consider other codecs and use cases where there is could be no mandatory ordering of headers defined by spec(but spec itself leveraging the responsibility to upper middleware)

QuPengfei:
agree@yakui @xhaihao, thanks to point out. the hack code will be improved with more generic method to check the nal type.

QuPengfei:
@sreerenjb  here is the order for the driver keep AUD/SPS/PPS/SEI/others/sliceheader/slicedata. one good point here is more robust when the middle insert them as the order of SPS/PPS/AUD, etc.

src/gen9_avc_encoder.c
```diff
@@ -6648,6 +6648,51 @@ gen9_mfc_avc_insert_object(VADriverContextP ctx,
...
+                                   encoder_context,
+                                   header_data,
+                                   ALIGN(length_in_bits, 32) >> 5,
+                                   length_in_bits & 0x1f,
+                                   skip_emul_byte_cnt,
+                                   0,
+                                   0,
+                                   !param->has_emulation_bytes,
+                                   0,
+                                   batch);
```
xhaihao: break if found

src/gen9_avc_encoder.c
```diff
@@ -6648,6 +6648,51 @@ gen9_mfc_avc_insert_object(VADriverContextP ctx,
...
+
+        param = (VAEncPackedHeaderParameterBuffer *)(encode_state->packed_header_params_ext[start_index + i]->buffer);
+
+        length_in_bits = param->bit_length;
+
+        skip_emul_byte_cnt = intel_avc_find_skipemulcnt((unsigned char *)header_data, length_in_bits);
+
+        /* skip the slice header packed data type as it is lastly inserted */
+        if (param->type != VAEncPackedHeaderRawData && (*(nal_type+skip_emul_byte_cnt-1)&0x1f) != AVC_NAL_DELIMITER )
+            continue;
```
xhaihao: As Yakui pointed out before, it is unnecessary to check the type against VAEncPackedHeaderRawData.

src/gen9_avc_encoder.c
QuPengfei: here the logic is not clear to read. i will change it.

src/gen9_avc_encoder.c
QuPengfei: here the logic is not clear to read. i will change it.

xhaihao:
@sreerenjb agree with you, ideally the driver should respect the submission order. Actually the driver won't change the order if the application submits all packed data via PackedHeaderRawData. As a fix to the current driver behavior, I'm fine with Pengfei's patch except a few changes

QuPengfei:
@xhaihao @sreerenjb it looks the driver's behavior more robus and help handle the wrong order in the middle ware.

QuPengfei:
update the patch

src/gen9_avc_encoder.c
```diff
@@ -6648,6 +6648,49 @@ gen9_mfc_avc_insert_object(VADriverContextP ctx,
...
+                                       encoder_context,
+                                       header_data,
+                                       ALIGN(length_in_bits, 32) >> 5,
+                                       length_in_bits & 0x1f,
+                                       skip_emul_byte_cnt,
+                                       0,
+                                       0,
+                                       !param->has_emulation_bytes,
+                                       0,
+                                       batch);
```
xhaihao: What I mean in the previous comment is the loop can break once a AUD nal is found and inserted at the beginning.

QuPengfei: sure

sreerenjb:
@QuPengfei Not really :), in a perfect scenario 1) driver follow the order which middleware submit the RawHeaders 2) Driver should have internal checking for the ordering  *only for the headers which the corresponding specification has mandatory precedence rule* and driver should warn/error out /hard_code_ordering if the application tried to break this mandatory rule.

But for now it could be okay to merge the patch

PR #125, opened 2017-04-06
gen9_avc_encoder: macro optimization

sreerenjb:
Avoid duplicated multiplication, do the frame size
calculation once and reuse the result.

Fixes #125 

Signed-off-by: Sreerenj Balachandran <sreerenj.balachandran@intel.com>


QuPengfei:
@sreerenjb good catch

PR #133, opened 2017-04-13
gen9_vp9_encoder: fix parameters sent to the shaders

### LINKED ISSUE #132: Update buffer fullness and initial distortion buffer information
VP9 encoder is not setting the proper buffer fullness as it has to be bitrate / 2.  Then the width and height in mb is not correct when building the shader information within intra_dist initialization.  This is causing that the first distortion buffer is all zeroed out and the history buffer uses that distortion buffer on the next inter frames.

This can be tested by analyzing an encoded stream and seeing that qp values for all frames keep going up.  This results in missing the bitrate targets on CBR mode.

chivakker:
buffer fullness intially has to be set to half the target bitrate.
Also the scaled 4x width and height in mb has to be updated properly
for distortion buffer to operate properly

Fixes #132
TEST="see bug for proper test analysis"


src/gen9_vp9_encoder.c
```diff
@@ -1076,6 +1076,7 @@ void gen9_vp9_set_curbe_brc(VADriverContextP ctx,
 
         *param->pbrc_init_reset_buf_size_in_bits  = cmd->dw6.buf_size;
         *param->pbrc_init_reset_input_bits_per_frame  = dInputBitsPerFrame;
+        *param->pbrc_init_current_target_buf_full_in_bits = cmd->dw6.buf_size >> 1;
```
yakuizhao: It will be better only to initialize it for init scenario instead of both init and reset scenario.

src/gen9_vp9_encoder.c
```diff
@@ -1365,8 +1366,8 @@ gen9_vp9_brc_intra_dist_kernel(VADriverContextP ctx,
     gen8_gpe_setup_interface_data(ctx, gpe_context);
 
     memset(&kernel_walker_param, 0, sizeof(kernel_walker_param));
-    kernel_walker_param.resolution_x = vme_context->downscaled_width_in_mb4x;
-    kernel_walker_param.resolution_y = vme_context->downscaled_height_in_mb4x;
+    kernel_walker_param.resolution_x = vp9_state->downscaled_width_4x_in_mb;
+    kernel_walker_param.resolution_y = vp9_state->downscaled_height_4x_in_mb;
```
yakuizhao: Good catch.

xhaihao: Could you remove downscaled_width_in_mb4x / downscaled_height_in_mb4x from vme_context? Those fields aren't used any more if applying your patch and it is easy to confuse reader if reserving these fields.

src/gen9_vp9_encoder.c
chivakker: This is needed on both init and reset scenarios

src/gen9_vp9_encoder.c
chivakker: That's a good suggestion I'll remove it from vme_context

Issue #114, opened 2017-04-04
gcc 4.8.5 reports possibly uninitialized variable in the driver


dvrogozh:
gcc 4.8.5 reports possibly uninitialized variable in the driver:
```
gen9_avc_encoder.c: In function 'gen9_avc_set_curbe_mbenc':
gen9_avc_encoder.c:3469:8: warning: 'cmd.g9' may be used uninitialized in this function [-Wmaybe-uninitialized]
     if(cmd.g9->dw4.use_actual_ref_qp_value)
```
Looking into the code, this happens in the gen9_avc_set_curbe_mbenc() function which is designed to be workable only on the SKL, BXT and KBL platforms. Please, make sure that the function has a default path for other platforms throwing assertion and error rather than trigger undefined behavior.

I was on the following commit:
```
commit 437cbe03eaeb08e6f4ab70cdf8757bccb8de15dd
Author: Víctor Manuel Jáquez Leal <vjaquez@igalia.com>
Date:   Mon Mar 27 09:05:40 2017 -0700

    gen8: accept P010 as valid format
    
    The functions gen8_gpe_set_surface2_state() and
    gen8_gpe_media_chroma_surface_setup() asserts for the surface color
    format, admiting only NV12.
    
    Nonetheless, gen8 also can encoder surfaces with P010 color format.
    
    This patch adds P010 color format in the assertion condition.
    
    Fixes #106
```

xhaihao:
This issue is gcc related. the issue disappears if you use a newer version of gcc. Yes, gen9_avc_set_curbe_mbenc was designed to be workable on SKL, BXT and KBL and presently gen9_avc_set_curbe_mbenc is only called on SKL, BXT and KBL. A newer version of gcc can handle this case.

yakuizhao:
Although the issue can be fixed on the newer GCC, it is still reasonable request that the code can be compiled/built correctly on the older GCC version.  The distribution system will use the specific GCC version. We can't ask them to update the GCC version  in order to fix one issue. 
So the reasonable fix is to make them be compatible with the old gcc version.

QuPengfei:
@dvrogozh this file gen9_avc_encoder.c is designed only for Gen9(SKL/KBL/APL). for other platform gen8/7 etc, the other path will be called.

xhaihao:
@yakuizhao the code can be built correctly with the older GCC

yakuizhao:
[...]

PR #136, opened 2017-04-13
correct coding style and silence a gcc warning 

xhaihao:
This fixes https://github.com/01org/intel-vaapi-driver/issues/114

### DIFF
--- a/src/i965_output_dri.c
+++ b/src/i965_output_dri.c
@@ -32,12 +32,9 @@
 
 #define LIBVA_X11_NAME "libva-x11.so.1"
 
-typedef struct dri_drawable *(*dri_get_drawable_func)(
-    VADriverContextP ctx, XID drawable);
-typedef union dri_buffer *(*dri_get_rendering_buffer_func)(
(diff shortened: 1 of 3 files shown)


src/i965_output_dri.c
```diff
@@ -35,7 +35,7 @@
 typedef struct dri_drawable *(*dri_get_drawable_func)(
     VADriverContextP ctx, XID drawable);
 typedef union dri_buffer *(*dri_get_rendering_buffer_func)(
-        VADriverContextP ctx, struct dri_drawable *d);
+            VADriverContextP ctx, struct dri_drawable *d);
 typedef void (*dri_swap_buffer_func)(
```
pengche1: why it has 8 blank space prefix? it is different with the above one.

xhaihao: @pengche1, It seems the code is indented incorrectly by the tool if the type of the returned value is union.  I will update the code to avoid this case.

xhaihao:
@pengche1, I updated the patches, could you help to review the patches again?

PR #139, opened 2017-04-19
Add the support of multi-slices for the HEVC encoder

pengche1:
Multi-slices hevc encoding is gotten back with this patch. It supports ARBITRARY_ROWS  and ARBITRARY_MBS slice structures


pengche1:
set attrib_list[i].value as VA_ENC_SLICE_STRUCTURE_ARBITRARY_MACROBLOCKS for it is not bit valid field in the va.h

src/i965_drv_video.c
```diff
@@ -1278,9 +1291,6 @@ i965_CreateConfig(VADriverContextP ctx,
                     attrib.value = 1;
                 else
                     attrib.value = I965_MAX_NUM_SLICE;
-            } else if (profile == VAProfileHEVCMain ||
-                       profile == VAProfileHEVCMain10) {
-                attrib.value = 1;
```
pengche1: it can be removed if avc multi-slices is also ready

xhaihao: No, it can't be removed even if avc multi-slices is ready on SKL+.  intel_enc_hw_context_init() will check whether VAConfigAttribEncMaxSlices is included in the config attribute list.

pengche1: sure, it will be updated

PR #148, opened 2017-04-28
ENC: fix the low quality at the first GOP for AVC encoder on the plat…

### LINKED ISSUE #147: Low quality at the first GOP for AVC encoder on the platform previous SKL

QuPengfei:
…form previous SKL

RC logic initialize the big QP at the first GOP, this cause the low quality.

Fixes #147


lizhong1008:
Would better the max_qp defined by middleware, instead of hard code in driver. I think we can add a max_qp in _VAEncMiscParameterRateControl just like staging branch.

QuPengfei:
max_qp is a feature in our next plan. this fix our RC logic initialization issue.

fhvwy:
This change seems odd - isn't the calculation immediately above exactly what is meant to be guessing the initial QP?  If that calculation is wrong, I think it be better to fix it rather than clipping the result afterwards to some arbitrary hard-coded range.

@lizhong1008  The initial_qp setting already exists for the user to set the QP for this specific case.  I agree that max_qp would also be desirable, but would only matter here when initial_qp is not supplied.

QuPengfei:
@fhvwy It is hard to caculate the intial QP for Genenral RC logic,especially for all the combination of resolution/bitrate/gopsize etc. here the similar initial experienced QP value (here are 35/40/45) also are  used in the other open source reference code with JVT-G012. The RC logic will begin to converge from the second GOP. The first GOP maybe a little bigger than average target bitrate especially under very low target bitrate.

For the max_qp, it is the common parameter for all frames (not seperate for I/P/B by now in the driver and not specified for a certain frame number). it is not good for this issue as it maybe need the middle to set it constantly back and forth.

src/gen6_mfc_common.c
```diff
@@ -188,9 +188,9 @@ static void intel_mfc_brc_init(struct encode_state *encode_state,
             mfc_context->brc.qp_prime_y[i][SLICE_TYPE_B] = mfc_context->brc.qp_prime_y[i][SLICE_TYPE_I];
         }
 
-        BRC_CLIP(mfc_context->brc.qp_prime_y[i][SLICE_TYPE_I], min_qp, 51);
-        BRC_CLIP(mfc_context->brc.qp_prime_y[i][SLICE_TYPE_P], min_qp, 51);
-        BRC_CLIP(mfc_context->brc.qp_prime_y[i][SLICE_TYPE_B], min_qp, 51);
+        BRC_CLIP(mfc_context->brc.qp_prime_y[i][SLICE_TYPE_I], min_qp, 36);
+        BRC_CLIP(mfc_context->brc.qp_prime_y[i][SLICE_TYPE_P], min_qp, 40);
+        BRC_CLIP(mfc_context->brc.qp_prime_y[i][SLICE_TYPE_B], min_qp, 45);
```
fhvwy: This additional clipping should not be applied if the user has supplied an initial_qp - maybe move it inside the immediately preceding block?

fhvwy:
I guess.  I still think that this is a hack and the numbers shouldn't be hard-coded here, but I don't have a any answer for that other than improving the QP decision logic.

In any case, this needs to be modified to not apply the hard-coded bound if initial_qp is set (see comment in code).

Wrt max_qp, yes I agree that it isn't very useful here.  It should be respected, but the existing initial_qp setting is the right place for the user to set this initial QP value if they want to.

src/gen6_mfc_common.c
fhvwy: Also BRC_CLIP() applies max before min, so it would do the wrong thing if min_qp is > 36.  Putting the new test inside the block and keeping the existing one outside would solve that too.

QuPengfei: @fhvwy are you talking about  pPicParameter->pic_init_qp in PPS?

thanks for point out the error when min_qp is > 36. it should be fixed. :)

QuPengfei:
here the patch only help to calulate the I/IDR frame QP in the first gop of the sequence and also it will be recaculated when context reset. it does not affect the initial QP setting in the PPS. this clip logic will avoid too big QP value in the first I frame and cause the big quality drop in the first GOP.

it is the common and practical trick in RC logic.

src/gen6_mfc_common.c
fhvwy: @QuPengfei  No.  I'm talking about encoder_context->brc.initial_qp, which is supplied by the user as VAEncMiscParameterRateControl.initial_qp.

QuPengfei: @fhvwy i see. the VAEncMiscParameterRateControl.initial_qp has the similar function as this patch.

PR #159, opened 2017-05-05
Fix: AVC encoder cause bad frame when quality level 1 on SKL/APL

### LINKED ISSUE #158: AVC encoder has the issue when quality level=1 on SKL/APL
quality level =1 cause the bad frame. this is regression issue.

QuPengfei:
Fixes #158


lizhong1008:
I'm fine for the fix code. But would better add more description message instead of "fix #158". The patch is fix the regression issue of BRC, not CQP. And also the regression commit can be provided.

Issue #35, opened 2017-02-09
Encoding doesn't check the input rate_control attribute 


yakuizhao:
When the upper-middleware passes the attribute that is not supported by the driver, the driver doesn't check it and still continue the encoding. This is wrong.
For example:  CBR is not exported. The upper-middleware passes the CBR  rate_control. And the driver still helps to do the CBR encoding. 

Expected behaviour:  return error code when the unsupported attribute is passed.

PR #169, opened 2017-05-16
ENC: add RC query for AVC mb rc mode on SKL+

### LINKED ISSUE #168: Add the query for MB RC mode for AVC on SKL+
this support the middle to query the RC mode for MB and use it properly

QuPengfei:
Fixes #168

### DIFF
--- a/src/i965_device_info.c
+++ b/src/i965_device_info.c
@@ -132,6 +132,8 @@ static struct hw_codec_info snb_hw_codec_info = {
     .has_tiled_surface = 1,
     .has_di_motion_adptive = 1,
 
+    .h264_brc_mode = VA_RC_CQP | VA_RC_CBR | VA_RC_VBR,
+
     .num_filters = 2,
     .filters = {
         { VAProcFilterNoiseReduction, I965_RING_NULL },
@@ -172,6 +174,8 @@ static struct hw_codec_info ivb_hw_codec_info = {
     .has_tiled_surface = 1,
     .has_di_motion_adptive = 1,
     .has_di_motion_compensated = 1,
+
+    .h264_brc_mode = VA_RC_CQP | VA_RC_CBR | VA_RC_VBR,
 
     .num_filters = 2,
     .filters = {
@@ -217,6 +221,8 @@ static struct hw_codec_info hsw_hw_codec_info = {
     .has_di_motion_compensated = 1,
     .has_h264_mvc_encoding = 1,
 
+    .h264_brc_mode = VA_RC_CQP | VA_RC_CBR | VA_RC_VBR,
+
     .num_filters = 5,
     .filters = {
         { VAProcFilterNoiseReduction, I965_RING_VEBOX },
@@ -261,6 +267,8 @@ static struct hw_codec_info bdw_hw_codec_info = {
     .has_di_motion_compensated = 1,
     .has_vp8_decoding = 1,
     .has_h264_mvc_encoding = 1,
+
+    .h264_brc_mode = VA_RC_CQP | VA_RC_CBR | VA_RC_VBR,
 
     .num_filters = 5,
     .filters = {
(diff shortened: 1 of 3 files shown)


src/i965_drv_video.c
```diff
@@ -974,12 +974,18 @@ i965_GetConfigAttributes(VADriverContextP ctx,
...
+
+                if (profile == VAProfileH264ConstrainedBaseline ||
                     profile == VAProfileH264Main ||
                     profile == VAProfileH264High ||
-                    profile == VAProfileHEVCMain)
-                    attrib_list[i].value |= VA_RC_VBR;
+                    profile == VAProfileHEVCMain) {
+                    if (IS_GEN9(i965->intel.device_info)) {
+                        attrib_list[i].value |= VA_RC_VBR;
+                        attrib_list[i].value |= VA_RC_MB;
```
lizhong1008: Can we declare the RC mode capacity in i965_device_info.c , just like lp_h264_brc_mode? It will make user have a unified interface to check codec capacity.

src/i965_drv_video.c
```diff
@@ -974,12 +974,18 @@ i965_GetConfigAttributes(VADriverContextP ctx,
                     attrib_list[i].value |= VA_RC_CBR;
 
                 if (profile == VAProfileVP8Version0_3 ||
-                    profile == VAProfileVP9Profile0 ||
-                    profile == VAProfileH264ConstrainedBaseline ||
+                    profile == VAProfileVP9Profile0)
+                    attrib_list[i].value |= VA_RC_VBR;
```
lizhong1008: Cherryview can support vp8 encoder, but without new kernel, right? So it should not support VBR.

xhaihao: Cherryview supports vp8 encoder with the new kernel. VBR is supported too.

lizhong1008: Ok, great to know this.

src/i965_drv_video.c
xhaihao: @QuPengfei what's your response to @lizhong1008's comment? I am fine with the current patch, but it would be great if you could do this.

QuPengfei:
@lizhong1008 @xhaihao sorry for late response. I will check it later. :)

src/i965_drv_video.c
QuPengfei: @lizhong1008 @xhaihao the variable lp_h264_brc_mode is an internal for driver and it is only used by query interface. the middle ware can check this by query. so i did not think of other use case by now.

QuPengfei: one use case is error handling for RC input checking in driver. One example is The middle user directly configure RC mode without querying the RC capability. vdenc encoder RC mode depend on the fw ready or not. so the potential issue is here.

QuPengfei: i saw the use case in https://github.com/01org/intel-vaapi-driver/issues/35

QuPengfei:
update the patch

src/i965_device_info.c
```diff
@@ -309,6 +317,8 @@ static struct hw_codec_info chv_hw_codec_info = {
     .has_h264_mvc_encoding = 1,
     .has_hevc_decoding = 1,
 
+    .h264_brc_mode = VA_RC_CQP | VA_RC_CBR,
```
xhaihao: @pengche1 a simple VBR was added on prev-skl by the commit below:

commit f804a502699c05b6a61afac833fd465df16b32c8
Author: Mark Thompson <sw@jkqxz.net>
Date:   Tue Jan 10 00:04:43 2017 +0000

H.264 encoder: add a simple reactive VBR rate control mode

QuPengfei: Is this feature in the release?

xhaihao: Yes, the patch was merged a few months ago.

PR #181, opened 2017-05-31
gen9_avc_encoder: Fix mbenc curbe size of KBL and GLK

sreerenjb:
Signed-off-by: Sreerenj Balachandran <sreerenj.balachandran@intel.com>

### DIFF
--- a/src/gen9_avc_encoder.c
+++ b/src/gen9_avc_encoder.c
@@ -4926,7 +4926,7 @@ gen9_avc_kernel_init_mbenc(VADriverContextP ctx,
         curbe_size = sizeof(gen9_avc_mbenc_curbe_data);
     } else if (IS_KBL(i965->intel.device_info) ||
                IS_GLK(i965->intel.device_info)) {
-        curbe_size = sizeof(gen9_avc_mbenc_curbe_data);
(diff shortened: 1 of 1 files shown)


src/gen9_avc_encoder.c
```diff
@@ -4926,7 +4926,7 @@ gen9_avc_kernel_init_mbenc(VADriverContextP ctx,
         curbe_size = sizeof(gen9_avc_mbenc_curbe_data);
     } else if (IS_KBL(i965->intel.device_info) ||
                IS_GLK(i965->intel.device_info)) {
-        curbe_size = sizeof(gen9_avc_mbenc_curbe_data);
+        curbe_size = sizeof(gen95_avc_mbenc_curbe_data);
     }
```
xhaihao: the mbenc curbe size of KBL/GLK used in gen9_avc_kernel_init_brc() is wrong too, could you fix it in the same commit?

sreerenjb:
@xhaihao  done!

xhaihao:
Thanks, sree.  It would be better to use the same PR when you update your commit in the future, if so, we can see the whole comments in the same PR.

PR #189, opened 2017-06-02
Set the gpe surface width of P010 as twice original pixel width

pengche1:
Signed-off-by: peng.chen <peng.c.chen@intel.com>

### DIFF
--- a/src/i965_gpe_utils.c
+++ b/src/i965_gpe_utils.c
@@ -2677,6 +2677,11 @@ gen9_add_2d_gpe_surface(VADriverContextP ctx,
     gpe_surface.cacheability_control = i965->intel.mocs_state;
     gpe_surface.format = format;
 
+    if (gpe_surface.is_media_block_rw) {
+        if (obj_surface->fourcc == VA_FOURCC_P010)
+            gpe_surface.is_16bpp = 1;
+    }
+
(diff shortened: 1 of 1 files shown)


src/i965_gpe_utils.c
```diff
@@ -1394,7 +1394,12 @@ i965_object_surface_to_2d_gpe_resource_with_align(struct i965_gpe_resource *res,
     unsigned int swizzle;
 
     res->type = I965_GPE_RESOURCE_2D;
-    res->width = ALIGN(obj_surface->orig_width, (1 << alignment));
+
+    if (obj_surface->fourcc == VA_FOURCC_P010)
+        res->width = ALIGN(obj_surface->orig_width * 2, (1 << alignment));
+    else
+        res->width = ALIGN(obj_surface->orig_width, (1 << alignment));
+
```
xhaihao: Usually the width / height in SURFACE_STATE is in pixel, not in byte, maybe you should change the surface format instead of the width / height.

pengche1: it should be based on the kernel implementation. it always need to set the format as 8bits even for P010

pengche1: No, the format is still set as 8bits even for P010 surface

pengche1:
new commit has been sent to this PR. please review it again. thanks!

PR #194, opened 2017-06-08
Fix: AVC/HEVC encoder quality range should be 7 on SKL+

### LINKED ISSUE #192: Fix the encoder quality range for AVC/HEVC on SKL+

QuPengfei:
Fixes #192


src/i965_drv_video.h
```diff
@@ -69,8 +69,8 @@
 #define DEFAULT_SATURATION      50
 
 #define ENCODER_QUALITY_RANGE     2
-#define ENCODER_QUALITY_RANGE_AVC     8
-#define ENCODER_QUALITY_RANGE_HEVC    8
+#define ENCODER_QUALITY_RANGE_AVC    7
+#define ENCODER_QUALITY_RANGE_HEVC   7
 #define ENCODER_DEFAULT_QUALITY   1
```
pengche1: for ENCODER_QUALITY_RANGE, the actual max input quality level is ENCODER_QUALITY_RANGE - 1. but this update let ENCODER_QUALITY_RANGE_AVC and ENCODER_QUALITY_RANGE_HEVC break this rule.

QuPengfei: quality level is defined as from 1 to quality range according to declaration

/**
     * \brief Encoding quality range attribute. Read-only.
     *
     * This attribute conveys whether the driver supports different quality level settings
     * for encoding. A value less than or equal to 1 means that the encoder only has a single
     * quality setting, and a value greater than 1 represents the number of quality levels
     * that can be configured. e.g. a value of 2 means there are two distinct quality levels.
     */
    VAConfigAttribEncQualityRange     = 21,

pengche1: agree with you. I has a misunderstanding of this macro.

PR #197, opened 2017-06-13
add check for brc mode when creating the config

### LINKED ISSUE #35: Encoding doesn't check the input rate_control attribute

QuPengfei:
Fixes #35


src/i965_drv_video.c
```diff
@@ -1281,6 +1281,18 @@ i965_CreateConfig(VADriverContextP ctx,
...
+        VAConfigAttrib attrib, *attrib_found;
+        attrib.type = VAConfigAttribRateControl;
+        attrib_found = i965_lookup_config_attribute(obj_config, attrib.type);
+        if ((entrypoint == VAEntrypointEncSlice) &&
+            !(attrib_found->value & i965->codec_info->h264_brc_mode))
+            vaStatus = VA_STATUS_ERROR_INVALID_CONFIG;
+        else if ((entrypoint == VAEntrypointEncSliceLP) &&
+                 !(attrib_found->value & i965->codec_info->lp_h264_brc_mode))
+            vaStatus = VA_STATUS_ERROR_INVALID_CONFIG;
+    }
```
xhaihao: It is possible the codec is not h264.

xhaihao:
@uartie, Could you help to check what is the error in CI machine?

src/i965_drv_video.c
QuPengfei: agree, other codec will have the issue.

uartie:
@xhaihao, it looks like there was a conflict in the container name that we used to compile this on CI.  Not sure how the conflict occurred.  I will go ahead and kick off another build.

QuPengfei:
update the patch

xhaihao:
I think we should do similar check for other vp8/vp9/hevc.

QuPengfei:
@xhaihao  agree to add for vp8/vp9/hevc. i suggest adding support in a separate patch.

xhaihao:
@QuPengfei #35 isn't a AVC specific issue. could you update the commit log if you want to add support for vp8/vp9/hevc in a separate patch?

QuPengfei:
update the commit message

PR #198, opened 2017-06-13
Wayland thread-safety

pkerling:
Fixes 01org/libva#11

It compiles, but I'm unsure how to test this. Do you know a simple program that uses the VAAPI wayland output which I could use for a test run?

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -16,6 +16,10 @@ m4_define([libva_package_version], [1.7.3])
 # libdrm minimum version requirement
 m4_define([libdrm_version], [2.4.52])
 
+# Wayland minimum version requirement
+# 1.11.0 for wl_proxy_create_wrapper
+m4_define([wayland_api_version], [1.11.0])
+
 AC_PREREQ([2.57])
 AC_INIT([intel_vaapi_driver],
         [intel_vaapi_driver_version],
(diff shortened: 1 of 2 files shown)


pkerling:
Also, why do you dynamically load everything from libwayland-client? Why don't you just link against it? It's a very light-weight library.

xhaihao:
@pkerling  putsurface_wayland is a simple program for wayland, you can find it in libva-utils.

xhaihao:
@pkerling  the intent is to drop explicit dependencies on wayland library, please check the commit beblow:

commit 41ceaf0fd179627960ae186f5fc3099e99157339
Author: Gwenole Beauchesne <gwenole.beauchesne@intel.com>
Date:   Mon Aug 6 16:11:47 2012 +0200

    wayland: drop explicit dependencies to Wayland libraries.

pkerling:
Great thanks, I'll test with putsurface_wayland then and report back.

Yes I can see that it drops the dependencies, but it provides no explanation/rationale. What's different from libva itself which already has a dependency on libwayland_client?

xhaihao:
The code was added a few years ago and I can't find the reason :(, but I agree with you that it should be fine to link against libwayland-client and simplify the code for wayland support.

pkerling:
Mh, that's a shame - most of the time there are of course reasons why it was done this or that way. But no one knows if they still apply if they have been lost to history anyway. So I guess I'll revert the old commit, fix the conflicts, and push it here, too. This way you can easily restore the dlopen stuff if you have to by just reverting one commit.

pkerling:
OK so I looked into it some more and `i965_output_dri.c` also uses dl-loading for `libva-x11.so`, so I guess the intention probably was to keep `i965_drv_video.so` free of direct references to any windowing systems. I'm unsure what to do here, but if this is changed then it should be done for both.

I checked the changes here with `putsurface_wayland` and it worked :) Would be cool if you could merge this, if the library loading is changed it should probably go into a new PR anyway.

PR #210, opened 2017-06-22
Remove generated file from tree and on clean

pkerling:
The Wayland protocol file is generated anyway, it should be removed from the tree.


xhaihao:
Could you add a Signed-off-by tag in the commit log ?

pkerling:
forgot about it again, sorry :( maybe you could add a commit hook that checks this?

xhaihao:
Yes, it would be better to add a check for this, I will try to add a hook.

xhaihao:
@uartie could you help to check what is blocking the build process?

PR #214, opened 2017-06-28
Remove useless frame dimension check for VP9

### LINKED ISSUE #213: VP9 encoder rejects width/height not divisible by 8
The codec supports arbitrary sizes - the MI blocks round the frame size up to the next multiple of 8.

fhvwy:
Fixes #213 - encoding with arbitrary sizes works on Kaby Lake after this change.


QuPengfei:
@YuJiankang does the media kernel have the limitation about widh/height? would you help double confirm?

PR #223, opened 2017-07-10
i965_surface_external_memory: calculate data_size when not provided

### LINKED ISSUE #222: vaCreateSurfaces asserts when using a external buffer
assert is on i965_drv_video.c:1443.

There's an assumption few lines before where height is calculated that data_size is not zero.  Client may not populate it but instead pitches and offsets.  When it is one plane that'll result in height being zero for obj_surface and the assert will be triggered

Please use pitches and offsets to calculate obj_surface->height properly

This is a major regression as no external buffer can be used.

chivakker:
Client may not provide total data_size, so it can be calculated
using pitches and width for obj_surface.

Fixes #222

### DIFF
--- a/src/i965_drv_video.c
+++ b/src/i965_drv_video.c
@@ -1483,7 +1483,12 @@ i965_suface_external_memory(VADriverContextP ctx,
         index >= memory_attibute->num_buffers)
         return VA_STATUS_ERROR_INVALID_PARAMETER;
 
-    obj_surface->size = memory_attibute->data_size;
+    if (!memory_attibute->data_size)
+        // no data_size provided by client, calculate it here
+        obj_surface->size = memory_attibute->height * memory_attibute->pitches[0];
+    else
+        obj_surface->size= memory_attibute->data_size;
+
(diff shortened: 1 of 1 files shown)


chivakker:
This patch is not working, needs to be revisited.

src/i965_drv_video.c
```diff
@@ -1483,7 +1483,12 @@ i965_suface_external_memory(VADriverContextP ctx,
         index >= memory_attibute->num_buffers)
         return VA_STATUS_ERROR_INVALID_PARAMETER;
 
-    obj_surface->size = memory_attibute->data_size;
+    if (!memory_attibute->data_size)
+        // no data_size provided by client, calculate it here
+        obj_surface->size = memory_attibute->height * memory_attibute->pitches[0];
```
sreerenjb: This is wrong. You should consider the pitches of all planes in order to calculate the obj_surface size.
Please have a look into the native_memory type implementation where the obj_surface size calculated correctly.

xhaihao:
The issue is fixed by another PR.

PR #229, opened 2017-07-19
Add Flexible Encoding Infrastructure (FEI) Support for Skylake

### LINKED ISSUE #228: Add Flexible Encoding Infrastructre(FEI) support in Skylake
FEI(Flexible Encoding Infrastructure) is an extension to VA API which allows applications to have more control over different phases of encoding and trade off quality for speed with their own IPs.

The main highlight of FEI is the possibility to split the encoding process into two phases, first is ENC and the second is PAK.ENC is the operation which performs all motion vector calculation and prediction.PAK is doing all transformations and entropy coding. Without having FEI, the whole ENC+PAK is a black box to middleware, but with FEI user can extract the output of ENC and feed PAK with a custom enhanced motion vectors and macroblock prediction modes.

VA-API(libva):
*************************
Patches are already landed in mailing list: https://github.com/01org/libva/pull/83
For now, we only have APIs for AVC encode

intel-vaapi-driver:
*************************
Skylake is the only platform which is supporting FEI now.
Patches are on the way.

sreerenjb:
Fixes #228


src/gen9_avc_encoder.c
```diff
@@ -3926,55 +4442,378 @@ gen9_avc_send_surface_mbenc(VADriverContextP ctx,
...
+        assert(size <= buffer_store->bo->size);
+        if (avc_priv_surface->res_fei_mb_cntrl_surface.bo != NULL)
+            i965_free_gpe_resource(&avc_priv_surface->res_fei_mb_cntrl_surface);
+        i965_dri_object_to_buffer_gpe_resource(
+            &avc_priv_surface->res_fei_mb_cntrl_surface,
+            buffer_store->bo);
+
+    } else if (fei_param->mb_input | fei_param->mb_size_ctrl) {
+        assert(fei_param->mb_ctrl != VA_INVALID_ID);
+    }
```
lizhong1008: Would better change it to: 
 if ((fei_param->mb_input | fei_param->mb_size_ctrl) {
 assert(fei_param->mb_ctrl != VA_INVALID_ID);
obj_buffer = BUFFER(fei_param->mb_ctrl);
...
}

src/gen9_avc_encoder.c
```diff
@@ -3926,55 +4442,378 @@ gen9_avc_send_surface_mbenc(VADriverContextP ctx,
...
+    struct gpe_media_object_walker_parameter media_object_walker_param;
+    struct gpe_encoder_kernel_walker_parameter kernel_walker_param;
+    unsigned int downscaled_width_in_mb, downscaled_height_in_mb;
+    int media_function = 0;
+    int kernel_idx = 0;
+    unsigned int mb_const_data_buffer_in_use = 0;
+    unsigned int mb_qp_buffer_in_use = 0;
+    unsigned int brc_enabled = 0;
+    unsigned int roi_enable = (generic_state->num_roi > 0) ? 1 : 0;
+    unsigned int dirty_roi_enable = ((generic_state->dirty_num_roi > 0) && (generic_state->frame_type == SLICE_TYPE_P) && (0));
```
lizhong1008: ((generic_state->dirty_num_roi > 0) && (generic_state->frame_type == SLICE_TYPE_P) && (0)) is always 0, why not let dirty_roi_enable = 0 directly?

src/i965_encoder.c
```diff
@@ -1453,6 +1468,16 @@ intel_enc_hw_context_init(VADriverContextP ctx,
             encoder_context->quality_range = ENCODER_QUALITY_RANGE_AVC;
         } else
             encoder_context->quality_range = ENCODER_QUALITY_RANGE;
+
+        if (obj_config->entrypoint == VAEntrypointFEI) {
+            encoder_context->fei_enabled = 1;
```
lizhong1008: Can disable fei_enabled when HAS_FEI_H264_ENCODING(i965) is not true, then not need to check Is_SKL() in gen9_avc_encode.c for FEI case.

src/gen9_avc_encoder.c
```diff
@@ -3596,6 +3662,456 @@ gen9_avc_set_curbe_mbenc(VADriverContextP ctx,
...
 
 static void
+gen9_avc_fei_set_curbe_mbenc(VADriverContextP ctx,
+                             struct encode_state *encode_state,
+                             struct i965_gpe_context *gpe_context,
+                             struct intel_encoder_context *encoder_context,
+                             void * param)
+{
+    struct i965_driver_data *i965 = i965_driver_data(ctx);
+    gen9_avc_fei_mbenc_curbe_data *cmd;
```
sreerenjb: Not much use. We anyway just map with i965_gpe_context_map_curbe() which is not checking whether map() was successful or not.

src/gen9_avc_encoder.c
```diff
@@ -3596,6 +3662,456 @@ gen9_avc_set_curbe_mbenc(VADriverContextP ctx,
...
+    struct encoder_vme_mfc_context * vme_context = (struct encoder_vme_mfc_context *)encoder_context->vme_context;
+    struct generic_enc_codec_state * generic_state = (struct generic_enc_codec_state *)vme_context->generic_enc_state;
+    struct avc_enc_state * avc_state = (struct avc_enc_state *)vme_context->private_enc_state;
+    VASurfaceID surface_id;
+    struct object_surface *obj_surface;
+    VAEncSliceParameterBufferH264 * slice_param = avc_state->slice_param[0];
```
sreerenjb: Why? both "ptr*" and "ptr *"are accepted representation and driver doesn't have a general rule for this kind of formatting.Both representations are heavily used all over the driver code.
May be Haihao can add more restrictions on "astyle xxx" for formatting :)

src/gen9_avc_encoder.c
```diff
@@ -3596,6 +3662,456 @@ gen9_avc_set_curbe_mbenc(VADriverContextP ctx,
...
+    unsigned int mbenc_i_frame_dist_in_use = curbe_param->mbenc_i_frame_dist_in_use;
+    unsigned int table_idx = 0;
+    int ref_width, ref_height, len_sp;
+    int is_bframe = (generic_state->frame_type == SLICE_TYPE_B);
+    unsigned int preset = generic_state->preset;
+
+    qp = pic_param->pic_init_qp + slice_param->slice_qp_delta;
+
+    cmd = (gen9_avc_fei_mbenc_curbe_data *)i965_gpe_context_map_curbe(gpe_context);
+    memset(cmd, 0, sizeof(gen9_avc_fei_mbenc_curbe_data));
```
sreerenjb: Again not much use unless we fix the i965_gpe_context_map_curbe() code to check whether drm_bo mapped successfully or not.

src/gen9_avc_encoder.c
```diff
@@ -3596,6 +3662,456 @@ gen9_avc_set_curbe_mbenc(VADriverContextP ctx,
...
+            break;
+        default:
+            assert(0);
+        }
+
+    }
+    /* 4 means full search, 6 means diamand search */
+    me_method  = (fei_param->search_window == 5) ||
+                 (fei_param->search_window == 8) ? 4 : 6;
+
```
sreerenjb: SeachWindow== 0 is the case where there is no predefined search window, and in this case, we have special index values for me method. But all other SEarchWindow values (1 to 8) have similar me method index calculations. So I would prefer the current implementation to avoid code duplication.

src/gen9_avc_encoder.c
```diff
@@ -3596,6 +3662,456 @@ gen9_avc_set_curbe_mbenc(VADriverContextP ctx,
...
+    }
+
+    /* ref_width*ref_height = Max 64x32 one direction, Max 32x32 two directions */
+    if (is_bframe) {
+        if (ref_width > 32) {
+            ref_width  = 32;
+        }
+        if (ref_height > 32) {
+            ref_height = 32;
+        }
```
sreerenjb: Yup, make sense. Will fix that. thanks

src/gen9_avc_encoder.c
```diff
@@ -3596,6 +3662,456 @@ gen9_avc_set_curbe_mbenc(VADriverContextP ctx,
...
+    }
+
```
sreerenjb: AFAIK, not needed. But I am not sure about this :)

src/gen9_avc_encoder.c
```diff
@@ -3596,6 +3662,456 @@ gen9_avc_set_curbe_mbenc(VADriverContextP ctx,
...
+    cmd->dw2.max_len_sp = len_sp;
+    cmd->dw38.max_len_sp = 0; // HLD mandates this field to be Zero
+    cmd->dw2.max_num_su = cmd->dw38.max_num_su = 57;
+    cmd->dw3.src_access =
+        cmd->dw3.ref_access = 0; // change it to (is_frame ? 0: 1) when interlace is suppoted
+
+    if (avc_state->ftq_enable && (generic_state->frame_type != SLICE_TYPE_I)) {
+        if (avc_state->ftq_override) {
+            cmd->dw3.ft_enable = avc_state->ftq_enable;
+
```
sreerenjb: Not really, is there any general rule like that? I mean "no space before <else> block or something similar". There are other places in driver where we put space to improve the readability, especially for long lines. eg: there is one in i965_GetConfigAttributes().

But I an remove those if there is a generl pattern like that :)

src/gen9_avc_encoder.c
```diff
@@ -3596,6 +3662,456 @@ gen9_avc_set_curbe_mbenc(VADriverContextP ctx,
...
+        } else {
+            if (generic_state->frame_type == SLICE_TYPE_P) {
+                cmd->dw3.ft_enable = gen9_avc_max_ftq_based_skip[preset] & 0x01;
+
```
sreerenjb: see the previous comment

src/gen9_avc_encoder.c
```diff
@@ -3596,6 +3662,456 @@ gen9_avc_set_curbe_mbenc(VADriverContextP ctx,
...
+    /* search path tables */
+    table_idx = (generic_state->frame_type == SLICE_TYPE_B) ? 1 : 0;
+    memcpy(&(cmd->dw16), table_enc_search_path[table_idx][me_method], 16 * sizeof(unsigned int));
+
+    //ftq_skip_threshold_lut set,dw14 /15
+
+    //r5 disable NonFTQSkipThresholdLUT
+    if (generic_state->frame_type == SLICE_TYPE_P) {
+        cmd->dw32.skip_val = gen9_avc_skip_value_p[avc_state->block_based_skip_enable][avc_state->transform_8x8_mode_enable][qp];
+
```
sreerenjb: see the previous comment

src/gen9_avc_encoder.c
```diff
@@ -3596,6 +3662,456 @@ gen9_avc_set_curbe_mbenc(VADriverContextP ctx,
...
+    } else if (generic_state->frame_type == SLICE_TYPE_B) {
+        cmd->dw32.skip_val = gen9_avc_skip_value_b[avc_state->block_based_skip_enable][avc_state->transform_8x8_mode_enable][qp];
+
```
sreerenjb: see the previous comment

src/gen9_avc_encoder.c
```diff
@@ -3596,6 +3662,456 @@ gen9_avc_set_curbe_mbenc(VADriverContextP ctx,
...
+    cmd->dw97.mad_surf_index = GEN9_AVC_MBENC_MAD_DATA_INDEX;
+    cmd->dw98.force_non_skip_mb_map_surface = GEN9_AVC_MBENC_FORCE_NONSKIP_MB_MAP_INDEX;
+    cmd->dw99.widi_wa_surf_index = GEN9_AVC_MBENC_WIDI_WA_INDEX;
+    cmd->dw100.brc_curbe_surf_index = GEN9_AVC_MBENC_BRC_CURBE_DATA_INDEX;
+    cmd->dw101.static_detection_cost_table_index = GEN9_AVC_MBENC_SFD_COST_TABLE_INDEX;
+    cmd->dw102.fei_mv_predictor_surf_index = GEN9_AVC_MBENC_MV_PREDICTOR_INDEX;
+    i965_gpe_context_unmap_curbe(gpe_context);
+
+    return;
+}
```
sreerenjb: The non-fei use case function for mbenc is even more bigger, 500+.
I would prefer to keep a single function since the code inside is not anything complicated, but rather setting values for different attributes.

src/gen9_avc_encoder.c
```diff
@@ -3926,55 +4442,378 @@ gen9_avc_send_surface_mbenc(VADriverContextP ctx,
...
-    struct i965_avc_encoder_context * avc_ctx = (struct i965_avc_encoder_context *)vme_context->private_enc_ctx;
     struct generic_enc_codec_state * generic_state = (struct generic_enc_codec_state *)vme_context->generic_enc_state;
     struct avc_enc_state * avc_state = (struct avc_enc_state *)vme_context->private_enc_state;
+    VAEncSliceParameterBufferH264 * slice_param = avc_state->slice_param[0];
+    VAEncMiscParameterFEIFrameControlH264 *fei_param = NULL;
+    struct object_buffer *obj_buffer = NULL;
+    struct buffer_store *buffer_store = NULL;
+    struct object_surface *obj_surface = NULL;
+    struct gen9_surface_avc *avc_priv_surface;
+    struct i965_gpe_resource *gpe_resource;
```
sreerenjb: avc_private_surface never get an unknown value because:
 if (!obj_surface || !obj_surface->private_data)
        return;

gpe_resource will always get the known value since it is assigned with an address of the already defined memory in avc_private_surface.

src/gen9_avc_encoder.c
sreerenjb: Yup, more clarity. I will do this. Thanks.

src/gen9_avc_encoder.c
sreerenjb: Not something I included in this patch series. I am trying to avoid any non-fei changes in this patch set to make it clean. If you need, please file a new bug on this.

src/i965_encoder.c
sreerenjb: Sorry, I don't understand???

sreerenjb:
@uartie Yup, I will try to add those in a separate patch. Thanks.

src/gen9_avc_encoder.c
lizhong1008: As the comments above: "/* ref_width*ref_height = Max 64x32 one direction, Max 32x32 two directions */". Seems the height needs to be clipped to 32, I guess, : )

src/i965_encoder.c
lizhong1008: I mean maybe we need to change it to:
if (obj_config->entrypoint == VAEntrypointFEI && HAS_FEI_H264_ENCODING(i965)) 
    encoder_context->fei_enabled = 1;
Then No need to check IS_SKL() in gen9_avc_encode.c, just like such code:
+        if (!encoder_context->fei_enabled) {
 +            generic_ctx->enc_kernel_ptr = (void *)skl_avc_encoder_kernels;
 +            generic_ctx->enc_kernel_size = sizeof(skl_avc_encoder_kernels);
 +        } else if (IS_SKL(i965->intel.device_info)) {
 +            generic_ctx->enc_kernel_ptr = (void *)skl_avc_fei_encoder_kernels;
 +            generic_ctx->enc_kernel_size = sizeof(skl_avc_fei_encoder_kernels);
 +        }
How do you think?

src/gen9_avc_encoder.c
lizhong1008: Yeah,  would better we can have a rule for the unified coding style.  Before Haihao do this, you can ignore my comment. : )

src/gen9_avc_encoder.c
sreerenjb: Yup, will fix it, thanks!

src/gen9_avc_encoder.c
```diff
@@ -3596,6 +3662,456 @@ gen9_avc_set_curbe_mbenc(VADriverContextP ctx,
...
+            i965_avc_get_max_mv_len(avc_state->seq_param->level_idc) * 4;//frame only
+
+        cmd->dw36.hme_combine_overlap = 1;
+
+        //check is_fwd_frame_short_term_ref
+        surface_id = slice_param->RefPicList1[0].picture_id;
+        obj_surface = SURFACE(surface_id);
+        if (!obj_surface) {
+            WARN_ONCE("Invalid backward reference frame\n");
+            return;
```
sreerenjb: Somehow I missed this comment :).
Good catch! But need a similar fix in a number of places which includes non-fei use cases too !
I will fix it for FEI use case in this patch.

src/i965_encoder.c
sreerenjb: If the code reaches intel_enc_hw_context_init() code block, the HAS_FEI_H264_ENCODING(i965) is always TRUE, because otherwise, config creation could have been already failed well ahead in vaCreateConfig.

But, one optimization is possible, which is to avoid the IS_SKL() check coming inside the (IS_SKL || IS_BXT) code block since fei_enabled == TRUE happens only for SKL.

sreerenjb:
Added Fixes based on previous reviews. Also, updated the media kernel binaries.

sreerenjb:
If anyone interested to test the FEI features, here is the gstreamer-vaapi implementatio:
https://cgit.freedesktop.org/~sree/gstreamer-vaapi/log/?h=fei-dev

sreerenjb:
@ceyusa No KBL unfortunately.

lizhong1008:
LGTM

xhaihao:
@sreerenjb could you fix the warning below:

> gen9_avc_encoder.c:6005:77: warning: array subscript is above array bounds [-Warray-bounds]
>          fei_misc_param = (VAEncMiscParameterBuffer*)encode_state->misc_param[VAEncMiscParameterTypeFEIFrameControl][0]->buffer;
>

src/gen9_avc_encoder.c
```diff
@@ -74,6 +75,14 @@
         OUT_BCS_BATCH(batch, attr);                             \
     } while (0)
 
+/* FEI specific buffer sizes per MB in bytes for gen9 */
+#define FEI_AVC_MB_CODE_BUFFER_SIZE      64
+#define FEI_AVC_MV_DATA_BUFFER_SIZE      128
+#define FEI_AVC_MB_CONTROL_BUFFER_SIZE   16
+#define FEI_AVC_MV_PREDICTOR_BUFFER_SIZE 40
+#define FEI_AVC_DISTORTION_BUFFER_SIZE   48
```
QuPengfei: if these defines are specified for Gen9 and different from Gen95 or later, can we add the prefix or postfix to identify them clearly

sreerenjb: These are basically equal to the corresponding VA structure size, so ideally should be similar across platforms. Anyway, we don't know whether g95 is going to bring any change in the required size, less likely though. I believe, Better keep the names as it is now and later we can change if needed.

QuPengfei: it is OK to me.

sreerenjb:
It seems the CI build failure is due to missing header file since it was not compiled against libva master which has FEI headers.

uartie:
@sreerenjb Travis CI will not trigger when libva dep changes.  Travis CI PR builds are only triggered when PR itself changes.  Intel CI triggers for both cases.

I'm not sure if you can force PR rebuild on Travis CI apart from just rebasing the PR.

uartie:
Looks like there is a "Restart Build" button on Travis CI page... I'll try it and see if it updates result.

uartie:
Yup, for Travis CI you just need to manually "Restart Build" to have it use latest libva master when it changes.

sreerenjb:
Thanks @uartie

xhaihao:
@sreerenjb I will bump the VA-API version to 0.40.1 as some new definitions were added,  could you should update the dependency once the new version is bumped?

xhaihao:
bumped a new VA API version in https://github.com/01org/libva/pull/105

sreerenjb:
Bumped the required libva version to 0.40.1

PR #232, opened 2017-07-21
Vdenc fixes

xhaihao:
Some fixes for VDEnc AVC encoding


xhaihao:
Add a commit to fix the type of brc_target_size in struct gen9_vdenc_context

Issue #13, opened 2017-01-29
[IVB/BYT/BSW] Macroblocking artifact seen on a VC1 video file


seanvk:
migrated from Bugzilla [#68492](https://bugs.freedesktop.org/show_bug.cgi?id=68492)
status REOPENED severity *major* in component *intel* for *---*
Reported in version *unspecified* on platform *All*
Assigned to: Focus.Luo

On 2013-08-23 23:01:04 +0000,  wrote:

> Macroblocking artifacts seen during playback of a WMV file with VC1 video using vaapi-decode. The macroblocks can be seen in the first 3-4 seconds of the clip. Also artifacts are more prominent around 02:31 timestamp in the playback.
> 
> Instructions to reproduce:
> 
> File: 
> http://download.microsoft.com/download/9/b/4/9b4535e2-f908-4785-97a5-ef360405e391/Elephant%27s%20Dream%20720p%202M.wmv
> 
> Playback command:
> gst-launch-0.10 filesrc location=elephant_dream.wmv ! asfdemux ! vaapidecode ! queue ! vaapisink
> 
> Platform: 
> reproduced on Ivybridge 64-bit and Baytrail-M 32-bit with 13.10 Ubuntu and 3.11 kernel.

On 2013-08-23 23:03:05 +0000,  wrote:

> These macroblocking artifacts are not seen when playing the same file with software codec, or with windows media player on Windows.

On 2014-10-01 15:58:39 +0000, Kevin Mitchell wrote:

> *** Bug 84552 has been marked as a duplicate of this bug. ***

On 2015-11-23 14:28:31 +0000, haihao wrote:

> We fixed some VC1 related issues this year. Could you try the latest driver if you still experience this issue ?

On 2016-01-12 04:48:16 +0000, Kevin Mitchell wrote:

> Problem persists unchanged for me with libva and intel-driver git master.

On 2016-02-05 20:04:27 +0000,  wrote:

> Gentlemen,
> 
> Seems like I'm having the same problem on ASRock Beebox Intel Celeron N3000. The file is coded VC-1, 1080p24. When I play the file with VAAPI enabled, I get the square artifacts on the screen. I experienced this problem with vaapi intel-driver 1.6.1 and 1.6.2. If VAAPI is disabled, the problem disappears, but the CPU load is too much.
> 
> You may get the file here: https://docs.google.com/uc?id=0B9X316MN6HHvSlNWb1ZhSGo2X2M&export=download (Second 3 till the end).
[...]

PR #234, opened 2017-07-25
implement intensity compensation for VC-1 decoding for Ivybridge and better

carpalis:
VC1-decoder: implement intensity compensation for VC-1

Intensity compensation was not present for B-frames, but only for P-frames. When a P-frame flags intensity compensation for its forward reference frame, all subsequent B-frames that use this reference frame as well, need to do intensity compensation as well.

Fixes: #13 [IVB/BYT/BSW] Macroblocking artifact seen on a VC1 video file

Signed-off-by: Jerome Borsboom <jerome.borsboom@carpalis.nl>

### DIFF
--- a/src/gen75_mfd.c
+++ b/src/gen75_mfd.c
@@ -1518,6 +1518,9 @@ gen75_mfd_init_vc1_surface(VADriverContextP ctx,
     }
 
     gen7_vc1_surface->picture_type = pic_param->picture_fields.bits.picture_type;
+    gen7_vc1_surface->intensity_compensation = 0;
+    gen7_vc1_surface->luma_scale = 0;
+    gen7_vc1_surface->luma_shift = 0;
 
     if (gen7_vc1_surface->dmv == NULL) {
         gen7_vc1_surface->dmv = dri_bo_alloc(i965->intel.bufmgr,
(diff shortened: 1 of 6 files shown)


xhaihao:
Could you please add a line of 'Fixes #xxx' to the commit message so that others know which issue is resolved by this commit?

src/gen75_mfd.c
```diff
@@ -1915,24 +1934,37 @@ gen75_mfd_vc1_pred_pipe_state(VADriverContextP ctx,
...
+    picture_type = pic_param->picture_fields.bits.picture_type;
+
+    if (gen7_mfd_context->reference_surface[0].surface_id != VA_INVALID_ID) {
+        if (picture_type == 1 || picture_type == 2) { /* P/B picture */
+            struct gen7_vc1_surface *gen7_vc1_surface = gen7_mfd_context->reference_surface[0].obj_surface->private_data;
+
+            intensitycomp_single_fwd = gen7_vc1_surface->intensity_compensation;
+            luma_scale1 = gen7_vc1_surface->luma_scale;
+            luma_shift1 = gen7_vc1_surface->luma_shift;
+        }
```
xhaihao: Does P frame also use the parameters from the reference frame? the commit message only says all subsequent B-frames need to do intensity compensation.

carpalis: The commit message only referenced the changes in intensity compensation. For P-frames intensity compensation was already working.

It is the P-frame that sets the parameters for its reference frame. Subsequently, the P-frame also uses this intensity compensated reference frame. Any B-frames that follow this P-frame and that reference that same reference frame also use the intensity compensation on the reference frame. (See 8.3.8 in VC-1 spec).

The patch stores the intensity compensation status with each surface. When such a surface is used as a reference, the stored intensity compensation status us used to program the graphics hardware.

xhaihao: thanks for the explanation in detail

Issue #252, opened 2017-08-18
gen8_post_processing: optimization for gen8_pp_object_walker


nelsoncs-intel:
gen8_pp_object_walker makes an unnecessary function call, checks a
return value that is always zero and makes non-indexed variable
assignments.  All of which occurs inside nested for loops.

Move non-indexed assignments outside the iterative block and inline the
rest. This decreases frame drops for saturating workloads.

[0001-gen8_post_processing-optimization-for-gen8_pp_object.txt](https://github.com/01org/intel-vaapi-driver/files/1235136/0001-gen8_post_processing-optimization-for-gen8_pp_object.txt)

chivakker:
@nelsoncs-intel can you please submit the patch as a pull request?

PR #253, opened 2017-08-18
gen8_post_processing: optimization for gen8_pp_object_walker

nelsoncs-intel:
gen8_pp_object_walker makes an unnecessary function call, checks a
return value that is always zero and makes non-indexed variable
assignments.  All of which occurs inside nested for loops.

Move non-indexed assignments outside the iterative block and inline the
rest. This reduces SOC power for modest workloads and decreases frame
drops for saturating workloads.

Fixes: #252

Signed-off-by: Cole Nelson <colex.nelson@intel.com>


xhaihao:
Could you please point out what is moved  outside the iterative block? The assignment to grf9 is still in the loop.

xhaihao:
line1427 and 1428 are a non-indexed variable assignment too,  could you move these lines outside the iterative block if you saw power saving in your testing although maybe the compiler does similar optimization.

nelsoncs-intel:
You are totally correct. 1427,1428 are non-indexed and I see another decrease in dropped frames under saturating load.  Currently retesting power.
As for optimization, I have disassembled a couple of times and not seen any 'silent' changes.  This testing is within the context of the ChromeOS build system, so other build settings/compilers could be optimizing differently.

xhaihao:
Could you squash the two commits into one commit?

uartie:
please rebase

montvid:
Why is this not merged?

uartie:
@xhaihao can you finish reviewing this PR, please?  It was submitted over a year ago.

PR #255, opened 2017-08-22
Gen8avc

laureatian:


src/i965_gpe_utils.c
```diff
@@ -2751,7 +2755,7 @@ gen9_add_buffer_2d_gpe_surface(VADriverContextP ctx,
     gpe_surface.cacheability_control = i965->intel.mocs_state;
     gpe_surface.format = format;
 
-    gen9_gpe_context_add_surface(gpe_context, &gpe_surface, index);
+    gpe->context_add_surface(gpe_context, &gpe_surface, index);
 }
```
xhaihao: Could you also replace the function name with a common name?

laureatian: Can I use i965 as the prefix of function name? The function only apply to gen8/9 and not apply to gen7 and before.

xhaihao: Yes. Actually the function can be applied to gen7 with few change.

src/gen9_avc_encoder.c
```diff
@@ -794,17 +929,17 @@ gen9_avc_allocate_resources(VADriverContextP ctx,
...
+    i965_free_gpe_resource(&avc_ctx->res_mb_status_buffer);
+    size = (generic_state->frame_width_in_mbs * generic_state->frame_height_in_mbs * 16 * 4 + 1023) & ~0x3ff;
+    allocate_flag = i965_allocate_gpe_resource(i965->intel.bufmgr,
+                                               &avc_ctx->res_mb_status_buffer,
+                                               ALIGN(size, 0x1000),
+                                               "MB statistics output buffer");
+    if (!allocate_flag)
+        goto failed_allocation;
+    i965_zero_gpe_resource(&avc_ctx->res_mb_status_buffer);
+    //}
```
QuPengfei: here we should keep the logic of "// if (avc_state->mb_status_supported) {" and set the "avc_state->mb_status_supported" accordlingly. comment out is a little confused and maybe affect the other platform logic

src/gen9_avc_encoder.c
```diff
@@ -1038,6 +1174,7 @@ gen9_avc_allocate_resources(VADriverContextP ctx,
                                                    "sfd output buffer");
         if (!allocate_flag)
             goto failed_allocation;
+        i965_zero_gpe_resource(&avc_ctx->res_sfd_output_buffer);
```
QuPengfei: Are the two "i965_zero_gpe_resource" necessary? it affect the performance as memory map operation.

src/gen9_avc_encoder.c
```diff
@@ -2595,26 +2865,118 @@ gen9_avc_send_surface_brc_frame_update(VADriverContextP ctx,
...
+    /* MB statistical data surface*/
+    gen9_add_buffer_2d_gpe_surface(ctx,
+                                   gpe_context,
+                                   &avc_ctx->res_mbbrc_mb_qp_data_surface,
+                                   1,
+                                   I965_SURFACEFORMAT_R8_UNORM,
+                                   GEN9_AVC_FRAME_BRC_UPDATE_MB_STATUS_INDEX);
+
+    return;
+}
```
QuPengfei: it looks gen8_avc_send_surface_brc_frame_update is not necessary and the better way is to reuse the function gen9_avc_send_surface_brc_frame_update.

src/gen9_avc_encoder.c
```diff
@@ -5376,9 +7041,11 @@ gen9_avc_encode_check_parameter(VADriverContextP ctx,
          rate_control_mode == VA_RC_CQP)) {
         generic_state->brc_need_reset = 0;// not support by CQP
     }
-
-    if (generic_state->brc_need_reset && !avc_state->sfd_mb_enable) {
+    // if (generic_state->brc_need_reset && !avc_state->sfd_mb_enable) {
```
QuPengfei: this line is commented out. how about remove it directly?

src/gen9_avc_encoder.c
```diff
@@ -4642,152 +5008,1442 @@ gen9_avc_kernel_wp(VADriverContextP ctx,
...
+        }
+    } else {
+        cmd->dw6.minimum_qp = 0 ;
+        cmd->dw6.maximum_qp = 0 ;
+    }
+    // cmd->dw6.enable_force_skip = avc_state->enable_force_skip ;
+    // cmd->dw6.enable_sliding_window = 0 ;
 
-    i965_gpe_context_unmap_curbe(gpe_context);
+    ge
```
QuPengfei: the function gen8_avc_kernel_init_brc is almost same as gen9_avc_kernel_init_brc. is it possible to reuse the gen9_avc_kernel_init_brc and just add some logic?

src/gen9_avc_encoder.c
```diff
@@ -4642,152 +5008,1442 @@ gen9_avc_kernel_wp(VADriverContextP ctx,
...
-    return VA_STATUS_SUCCESS;
+    return VA_STATUS_SUCCESS;
+}
+
+
+/*
+sfd related function
+*/
+static void
+gen9_avc_set_curbe_sfd(VADriverContextP ctx,
```
QuPengfei: Just confused. i did not see the code changes in this function gen9_avc_set_curbe_sfd. why it is also showed in the patch?
Same as gen9_avc_send_surface_sfd/gen9_avc_kernel_sfd?

src/gen9_avc_encoder.c
```diff
@@ -4642,152 +5008,1442 @@ gen9_avc_kernel_wp(VADriverContextP ctx,
...
+    curbe_cmd->dw36.vme_fwd_inter_pred_surf_index = GEN8_AVC_ME_CURR_FOR_FWD_REF_CM;
+    curbe_cmd->dw37.vme_bdw_inter_pred_surf_index = GEN8_AVC_ME_CURR_FOR_BWD_REF_CM;
+    curbe_cmd->dw38.reserved = 0;
+
+    i965_gpe_context_unmap_curbe(gpe_context);
+    return;
+}
+
+static void
+gen8_avc_send_surface_me(VADriverContextP ctx,
```
QuPengfei: is it possible to reuse the gen9_avc_send_surface_me? it looks almost the same.

src/gen9_avc_encoder.c
laureatian: if avc_ctx->res_sfd_output_buffer  is not set to zeros, It will cause inconsistent performance.

PR #259, opened 2017-08-24
Fix deprecated warning

xhaihao:
enums and VA features marked as deprecated should not be used any more.

### DIFF
--- a/src/i965_decoder_utils.c
+++ b/src/i965_decoder_utils.c
@@ -962,12 +962,9 @@ intel_decoder_check_avc_parameter(VADriverContextP ctx,
     ASSERT_RET((pic_param->CurrPic.picture_id != VA_INVALID_SURFACE), VA_STATUS_ERROR_INVALID_PARAMETER);
     ASSERT_RET((pic_param->CurrPic.picture_id == decode_state->current_render_target), VA_STATUS_ERROR_INVALID_PARAMETER);
 
-    if ((h264_profile != VAProfileH264Baseline)) {
-        if (pic_param->num_slice_groups_minus1 ||
-            pic_param->pic_fields.bits.redundant_pic_cnt_present_flag) {
-            WARN_ONCE("Unsupported the FMO/ASO constraints!!!\n");
-            goto error;
-        }
+    if (pic_param->pic_fields.bits.redundant_pic_cnt_present_flag) {
+        WARN_ONCE("Unsupported the ASO constraints!!!\n");
+        goto error;
     }
 
     /* Fill in the reference objects array with the actual VA surface
--- a/src/i965_drv_video.h
+++ b/src/i965_drv_video.h
@@ -40,6 +40,8 @@
 #include <va/va_backend.h>
 #include <va/va_backend_vpp.h>
 
+#include "va_backend_compat.h"
+
 #include "i965_mutext.h"
 #include "object_heap.h"
 #include "intel_driver.h"
--- a/src/va_backend_compat.h
+++ b/src/va_backend_compat.h
@@ -58,4 +58,12 @@
 
 #endif
 
+#if VA_CHECK_VERSION(1,0,0)
+
+# define VAEncPackedHeaderMiscMask      0x80000000
+# define VAEncPackedHeaderH264_SEI      (VAEncPackedHeaderMiscMask | 1)
+# define VAEncPackedHeaderHEVC_SEI      (VAEncPackedHeaderMiscMask | 1)
+
+#endif
+
 #endif /* VA_BACKEND_COMPAT_H */
--- a/test/Makefile.am
+++ b/test/Makefile.am
@@ -102,7 +102,6 @@ test_i965_drv_video_CPPFLAGS =						\
 
 test_i965_drv_video_CXXFLAGS =						\
 	-Wall -Werror							\
-	-Wno-deprecated-declarations					\
 	$(AM_CXXFLAGS)							\
 	$(NULL)
 
--- a/test/i965_avcd_config_test.cpp
+++ b/test/i965_avcd_config_test.cpp
@@ -96,8 +96,6 @@ VAStatus HasMVCDecodeSupport()
 }
 
 static const std::vector<ConfigTestInput> inputs = {
-    { VAProfileH264Baseline, VAEntrypointVLD, &ProfileNotSupported },
-
     { VAProfileH264ConstrainedBaseline, VAEntrypointVLD, &HasDecodeSupport },
     { VAProfileH264Main, VAEntrypointVLD, &HasDecodeSupport },
     { VAProfileH264High, VAEntrypointVLD, &HasDecodeSupport },
--- a/test/i965_avce_config_test.cpp
+++ b/test/i965_avce_config_test.cpp
@@ -129,10 +129,6 @@ VAStatus HasMVCEncodeSupport()
 }
 
 static const std::vector<ConfigTestInput> inputs = {
-    {VAProfileH264Baseline, VAEntrypointEncSlice, &ProfileNotSupported},
-    {VAProfileH264Baseline, VAEntrypointEncSliceLP, &ProfileNotSupported},
-    {VAProfileH264Baseline, VAEntrypointEncPicture, &ProfileNotSupported},
-
     {VAProfileH264ConstrainedBaseline, VAEntrypointEncSlice, &HasEncodeSupport},
     {VAProfileH264ConstrainedBaseline, VAEntrypointEncSliceLP, &HasLPEncodeSupport},
     {VAProfileH264ConstrainedBaseline, VAEntrypointEncPicture, &H264NotSupported},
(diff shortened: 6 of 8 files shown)


src/i965_decoder_utils.c
```diff
@@ -962,14 +962,6 @@ intel_decoder_check_avc_parameter(VADriverContextP ctx,
     ASSERT_RET((pic_param->CurrPic.picture_id != VA_INVALID_SURFACE), VA_STATUS_ERROR_INVALID_PARAMETER);
     ASSERT_RET((pic_param->CurrPic.picture_id == decode_state->current_render_target), VA_STATUS_ERROR_INVALID_PARAMETER);
 
-    if ((h264_profile != VAProfileH264Baseline)) {
-        if (pic_param->num_slice_groups_minus1 ||
```
uartie: Is this check not needed anymore for the other AVC profiles?

xhaihao: Other AVC profiles only supports single slice group (num_slice_groups_minus1 = 0),  so if  num_slice_groups_minus1 is not 0 in the stream.  it is not a constrained baseline / main / high profile stream and application should not use VA-API for this stream.

uartie: Ok, so if the application happens to use VAAPI for such stream (constrained baseline / main / high profile) wouldn't we still want to provide this warning if num_slice_groups_minus1 is not 0?  Or is this constraint dealt with at another level?

uartie: Maybe I'm stating the obvious... but the outer conditional is **true** for all **non-deprecated** avc profiles, and therefore the inner conditional is applied for such profiles.  Removing the inner conditional changes functionality for the remaining **non-deprecated** avc profiles.

xhaihao: For constrainedbaseline / main / high profile, num_slice_groups_minus1 must be 0. Do you mean application take a baseline stream (num_slice_groups_minus1 is set to non-zero) as a constrainedbaseline / main / high profile and use VAAPI for this stream no matter the value of num_slice_groups_minus1? I think it is hard for driver to avoid such usage if application doesn't follow the spec to use VAAPI even if we don't remove this check, e.g.  application set num_slice_groups_minus1 to 0 for a baseline stream, how does the driver know is a baseline stream, not a contrained baseline stream?

uartie: Say, for example, the following values are set upon function entry:

```
h264_profile = VAProfileH264Main
pic_param->num_slice_groups_minus1= 1
```

With the above hypothetical values, the code block in the innermost conditional would get executed.  That is, a warning would be emitted and then immediately transfer control to the "error" section, which returns VA_STATUS_ERROR_INVALID_PARAMETER to the application.

Now, if this section of code is removed and we have the above hypothetical values, control would no longer transfer to the "error" section but instead continue as if normal.  Therefore functionality is changed (either for better or worse).

Of course, I'm assuming that user/app has ability and freedom to set the above hypothetical values (whether it follows spec or not).  Please correct me if I'm wrong.

uartie: Perhaps it's better to demonstrate my point with a unit test.

Copy the below unit test case code into, say, the `test/i965_avcd_config_test.cpp` file and compile driver (--enable-tests) **without** your patches.
```
class AVCDecodeTest : public I965TestFixture { };

TEST_F(AVCDecodeTest, VAProfileH264Main_InvalidNumSliceGroupsMinus1)
{
    VAConfigAttrib a = { type:VAConfigAttribRTFormat, value:VA_RT_FORMAT_YUV420 };
    ConfigAttribs attribs(1, a);

    VAPictureParameterBufferH264 pparam;
    pparam.num_slice_groups_minus1 = 1;

    VASliceParameterBufferH264 sparam;
    std::vector<uint8_t> data = {0x00};

    ASSERT_NO_FAILURE(Surfaces surfaces = createSurfaces(1, 1, VA_RT_FORMAT_YUV420));

    pparam.CurrPic.picture_id = surfaces.front();

    ASSERT_NO_FAILURE(VAConfigID config = createConfig(VAProfileH264Main, VAEntrypointVLD, attribs));
    ASSERT_NO_FAILURE(VAContextID context = createContext(config, 1, 1, 0, surfaces));
    ASSERT_NO_FAILURE(VABufferID sliceDataBufId = createBuffer(context, VASliceDataBufferType, 1, 1, data.data()));
    ASSERT_NO_FAILURE(VABufferID sliceParamBufId = createBuffer(context, VASliceParameterBufferType, sizeof(sparam), 1, &sparam));
    ASSERT_NO_FAILURE(VABufferID picBufId = createBuffer(context, VAPictureParameterBufferType, sizeof(pparam), 1, &pparam));

    ASSERT_NO_FAILURE(beginPicture(context, surfaces.front()));
    ASSERT_NO_FAILURE(renderPicture(context, &picBufId));
    ASSERT_NO_FAILURE(renderPicture(context, &sliceParamBufId));
    ASSERT_NO_FAILURE(renderPicture(context, &sliceDataBufId));
    ASSERT_STATUS_EQ(i965_EndPicture(*this, context), VA_STATUS_ERROR_INVALID_PARAMETER);
}
```
Then execute:
```
test/test_i965_drv_video --gtest_filter=AVCDecodeTest.VAProfileH264Main_InvalidNumSliceGroupsMinus1
```
I get the following output:
```
Note: Google Test filter = AVCDecodeTest.VAProfileH264Main_InvalidNumSliceGroupsMinus1
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
Seeded std::rand() with 1503637661.
libva info: VA-API version 1.0.0
libva info: va_getDriverName() returns 0
libva info: User requested driver 'i965'
libva info: Trying to open /home/uaeoff/Work/workspace/media/build/intel-vaapi-driver/src/.libs/i965_drv_video.so
libva info: Found init function __vaDriverInit_1_0
libva info: va_openDriver() returns 0
[----------] 1 test from AVCDecodeTest
[ RUN      ] AVCDecodeTest.VAProfileH264Main_InvalidNumSliceGroupsMinus1
WARNING: Unsupported the FMO/ASO constraints!!!
[       OK ] AVCDecodeTest.VAProfileH264Main_InvalidNumSliceGroupsMinus1 (0 ms)
[----------] 1 test from AVCDecodeTest (0 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (1 ms total)
[  PASSED  ] 1 test.
```
You can see we've hit the warning condition and immediately returned to application/test with expected status.

Now, apply your patches, compile and run the test again.  I get the following output:

```
Note: Google Test filter = AVCDecodeTest.VAProfileH264Main_InvalidNumSliceGroupsMinus1
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
Seeded std::rand() with 1503637834.
libva info: VA-API version 1.0.0
libva info: va_getDriverName() returns 0
libva info: User requested driver 'i965'
libva info: Trying to open /home/uaeoff/Work/workspace/media/build/intel-vaapi-driver/src/.libs/i965_drv_video.so
libva info: Found init function __vaDriverInit_1_0
libva info: va_openDriver() returns 0
[----------] 1 test from AVCDecodeTest
[ RUN      ] AVCDecodeTest.VAProfileH264Main_InvalidNumSliceGroupsMinus1
i965_avcd_config_test.cpp:141: Failure
      Expected: VaapiStatus(i965_EndPicture(*this, context))
      Which is: VA_STATUS_ERROR_INVALID_SURFACE
To be equal to: VaapiStatus(0x00000012)
      Which is: VA_STATUS_ERROR_INVALID_PARAMETER
[  FAILED  ] AVCDecodeTest.VAProfileH264Main_InvalidNumSliceGroupsMinus1 (0 ms)
[----------] 1 test from AVCDecodeTest (0 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (1 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] AVCDecodeTest.VAProfileH264Main_InvalidNumSliceGroupsMinus1

 1 FAILED TEST
```
Obviously the surface is invalid and likely other settings,  but this is for demonstration purpose only.  Consider if surface and other params are valid (except num_slice_groups_minus1), then other unintentional things could possibly happen.  But I'm not actually an expert in avc decoding so it's really up to you to decide risk here.

Hopefully I'm not being overzealous ;)

xhaihao: I understand your concern, but num_slice_groups_minus1 was marked as deprecated, which means application should not use it any more and num_slice_groups_minus1 might be uninitialized at all.

uartie: Well that's embarrassing... I did not realize num_slice_groups_minus1 was deprecated!  You are right.

With that settled let's move on to the second part of the inner conditional:
```
 || pic_param->pic_fields.bits.redundant_pic_cnt_present_flag
```
... all the same previous concerns apply again if `redundant_pic_cnt_present_flag` != 0

xhaihao:
@uartie , I updated the patch series, could you please review again?

PR #261, opened 2017-08-31
Add several asserts to detect unvalid NULL pointer

pengche1:
Signed-off-by: peng.chen <peng.c.chen@intel.com>

### DIFF
--- a/src/gen9_avc_encoder.c
+++ b/src/gen9_avc_encoder.c
@@ -3691,6 +3691,7 @@ gen9_avc_fei_set_curbe_mbenc(VADriverContextP ctx,
 
     qp = pic_param->pic_init_qp + slice_param->slice_qp_delta;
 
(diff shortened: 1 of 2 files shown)


src/gen9_avc_encoder.c
```diff
@@ -4016,7 +4017,7 @@ gen9_avc_fei_set_curbe_mbenc(VADriverContextP ctx,
         obj_surface = SURFACE(surface_id);
         if (!obj_surface) {
             WARN_ONCE("Invalid backward reference frame\n");
-            if (gpe_context)
+            if (gpe_context != NULL)
```
xhaihao: I think this change is unnecessary .  the two conditions are the same indeed

pengche1: ok, I will delete this update

PR #271, opened 2017-09-18
Change the alignment for height  for linear surface

xhaihao:
The memory is allocated together for planar YUV, but we may use
Y, U, V surfaces for 3-plane YUV or Y, UV surfaces for 2-plane YUV
separately, e.g. a NV12 surface is taken as 2 separate surfaces by
the implemetation of vaPutSurface() in the driver. The memory format
for each surface is either linear or tiled. In the past we thought linear
surface requires 4K aligned surface base address, like as tiled surface,
so we set the alighment for height to 16. Actually linear surface only
requires base address is naturally-aligned to the element size. In
addition, some operations in the driver requires surface height is a
multiple of 4. so we can set the alignment for height to 4 for linear
surface.

### DIFF
--- a/src/i965_device_info.c
+++ b/src/i965_device_info.c
@@ -72,7 +72,7 @@ static struct hw_codec_info g4x_hw_codec_info = {
     .max_width = 2048,
     .max_height = 2048,
     .min_linear_wpitch = 16,
-    .min_linear_hpitch = 16,
+    .min_linear_hpitch = 4,
 
     .has_mpeg2_decoding = 1,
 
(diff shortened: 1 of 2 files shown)


chivakker:
@xhaihao I have used this patch to run some validation and it seems to be working fine.  Although there is a chance of an assert to be hit on the driver at i965_surface_external_memory function

    if (memory_attibute->num_planes > 1) {
        ASSERT_RET(IS_ALIGNED(obj_surface->height, 16), VA_STATUS_ERROR_INVALID_PARAMETER);
        ASSERT_RET(obj_surface->height >= obj_surface->orig_height, VA_STATUS_ERROR_INVALID_PARAMETER);
    }

Can you explain how this is suppose to work for linear surfaces?  AFAICT, when having a 3 plane external surface which is allocated 4 bit aligned, will hit this assert for height being 16 bit aligned.  Could it be relaxed as well?  It seems like so based on some testing although I want to understand possible implications.

xhaihao:
@chivakker The check against 16 is wrong, we should remove this ASSERT_RET().

sreerenjb:
> . In
> addition, some operations in the driver requires surface height is a
> multiple of 4
@xhaihao  Just curious, which operation requires this specific restriction?

PR #280, opened 2017-10-01
Remove unused variable

fhvwy:


xhaihao:
I see some other unused variables when using a newer version of gcc, could you help to remove these variables as well in this PR?

fhvwy:
@xhaihao  Updated, thanks.

PR #287, opened 2017-10-22
Fix common csc scaling

xhaihao:
A patch series to enhance the common path for CSC and scaling.


src/gen8_post_processing.c
```diff
@@ -328,6 +328,28 @@ static const uint32_t pp_yuv420p8_scaling_gen8[][4] = {
...
+        "8bit to 8bit",
+        0,
+        pp_yuv420p8_scaling_gen8,
+        sizeof(pp_yuv420p8_scaling_gen8),
+        NULL,
+    },
+
+    {
+        "8bit 420 to rgb32",
+        1,
```
pengche1: how about 10bit 420 to rgb32?

xhaihao: the legacy path is still used. Of course we can add support for 10bit 420 to rgb32 in the future.

xhaihao:
@pengche1 @QuPengfei  Could you help to review this PR?

PR #292, opened 2017-11-06
add meson support

ceyusa:
add meson support, cleaning up some partos of the autotools support


sreerenjb:
Nice work! Thanks, Victor :)

uartie:
What about the "test" directory?

ceyusa:
> What about the "test" directory?

Oops. I guess this is a first iteration then. Next step, the tests.

ceyusa:
I've added a bunch of fixes for autoools, mostly removing not used checks and enabling the out-of-source building.

In the case of meson, now it builds the driver from the static library, but for that it is required meson > 0.43

And the tests also uses that static library, just as in autotools, nonetheless the test fails, because the struct i965_driver_data get some how partially overwritten when it is cast in the test fixture. Quite strange.

test/meson.build
```diff
@@ -0,0 +1,76 @@
...
+
+gtest_includes = [
+  include_directories('gtest'),
+  include_directories('gtest/include'),
+]
+
+libgtest = static_library(
+  'gtest',
+  gtest_sources,
+  include_directories : gtest_includes)
```
uartie: `dependencies : thread_dep`?  Autotools has it implicit in AM_LDFLAGS.

c++11 std? Autotools has it implicit in AM_CPPFLAGS

test/meson.build
```diff
@@ -0,0 +1,76 @@
...
+  include_directories('gtest/include'),
+]
+
+testdriverdir = join_paths(meson.build_root(), 'src')
+cpp_flags = [
+  '-DTEST_VA_DRIVERS_PATH="' + testdriverdir + '"',
+  '-DGTEST_HAS_PTHREAD=' + thread_dep.found().to_string('1', '0'),
+  '-DGTEST_LANG_CXX11=1',
+  '-DGTEST_USE_OWN_TR1_TUPLE=0',
+  '-DGTEST_HAS_TR1_TUPLE=1',
```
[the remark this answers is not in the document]
ceyusa: > Also, -DPTHREAD?

In src's meson.build there is

```
if thread_dep.found()
  cflags += [ '-DPTHREADS' ]
endif
```

test/meson.build
ceyusa: > c++11 std? Autotools has it implicit in AM_CPPFLAGS

In test's meson.build:

```
  include_directories : gtest_includes,
  override_options : [ 'cpp_std=c++11' ])
```

test/meson.build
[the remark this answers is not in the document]
uartie: Right, but cflags aren't added to any of test's meson.build targets.  If you turn on verbose compile, then you'll see that it is missing during compile of test targets.

test/meson.build
uartie: Yes, but it is only specified in the test_i965 target and not the libgtest target.  It needs to be in both targets.

uartie:
-DHAVE_CONFIG_H is also missing from test_i965 target.

uartie:
VA_DRIVERS_PATH is used in i965_drv_video.c, thus need to add -DVA_DRIVERS_PATH=... to meson configuration.

test/meson.build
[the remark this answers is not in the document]
uartie: -DPTHREADS is definitely needed since without it the sizeof(_I965Mutex) will not be the same between i965_drv_video.so and test_i965_drv_video binaries.  This will effect the sizeof(struct object_heap) and sizeof(struct i965_driver_data) also.  Hence, the reason why you see strange behavior with i965_driver_data in tests.  That is, vaInitialize will dlopen the driver, which has its own compiled-in struct definition whereas the test fixture will use its own compiled-in struct definition.

In general, both driver and test executable should always be compiled with exact same compiler defines.  More specifically, any preprocessor definitions on the compiler command line (and config.h) that get used in any of the headers must match for both targets to ensure exact same data structures are defined.

test/meson.build
ceyusa: > Yes, but it is only specified in the test_i965 target and not the libgtest target. It needs to be in both targets.

Indeed.

test/meson.build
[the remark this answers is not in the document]
ceyusa: As far as I understand, PTHREADS macro is only needed by libi965_drv_video static library and by the driver it self, so, if I understand correctly, there's no need to set it for the tests since they are linking the already compiled library.

ceyusa:
> -DHAVE_CONFIG_H is also missing from test_i965 target.

Yes! I overlooked that. Thanks

> VA_DRIVERS_PATH is used in i965_drv_video.c, thus need to add -DVA_DRIVERS_PATH=... to meson configuration.

I think this is the same case of PTHREADS. I might be wrong, but afaik, it is already compiled in the library.

ceyusa:
I was wrong: PTHREADS is its required to link the library

ceyusa:
just found an issue when compiling as a subproject (gst-build, for example)

I'll push another pull-request in a moment

test/meson.build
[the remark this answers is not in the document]
uartie: PTHREADS is used in driver headers and the tests include driver headers.

uartie:
Please separate the autotools patches into a different pull-request.

uartie:
>> VA_DRIVERS_PATH is used in i965_drv_video.c, thus need to add -DVA_DRIVERS_PATH=... to meson configuration.

> I think this is the same case of PTHREADS. I might be wrong, but afaik, it is already compiled in the library.

VA_DRIVERS_PATH is missing in the library target.

test/meson.build
```diff
@@ -0,0 +1,84 @@
...
+  include_directories('gtest'),
+  include_directories('gtest/include'),
+]
+
+cpp_flags = [
+  '-DGTEST_HAS_PTHREAD=' + thread_dep.found().to_string('1', '0'),
+  '-DGTEST_LANG_CXX11=1',
+  '-DGTEST_USE_OWN_TR1_TUPLE=0',
+  '-DGTEST_HAS_TR1_TUPLE=1',
+]
```
uartie: Please add these to the test_i965_drv_video target, too, since they are used in gtest headers included by both targets.

src/meson.build
```diff
@@ -0,0 +1,241 @@
+config_cfg = configuration_data()
+config_cfg.set('VERSION', intel_vaapi_driver_version)
+config_cfg.set('VA_DRIVER_INIT_FUNC', va_driver_init_func)
+config_cfg.set('INTEL_DRIVER_MAJOR_VERSION', intel_vaapi_driver_major_version)
+config_cfg.set('INTEL_DRIVER_MINOR_VERSION', intel_vaapi_driver_minor_version)
+config_cfg.set('INTEL_DRIVER_MICRO_VERSION', intel_vaapi_driver_micro_version)
+config_cfg.set('INTEL_DRIVER_PRE_VERSION', intel_vaapi_driver_pre_version)
+config_cfg.set10('HAVE_HYBRID_CODEC', get_option('enable_hybrid_codec'))
+config_cfg.set10('HAVE_VA_X11', not get_option('disable_x11'))
```
uartie: Need to check for libva-x11... otherwise compilation fails if libva-x11 is missing.  Better to fail during configure.

meson.build
```diff
@@ -0,0 +1,86 @@
...
+va_driver_init_func = '__vaDriverInit_@0@_@1@'.format(va_api_major_version,
+                                                      va_api_minor_version)
+
+WITH_WAYLAND = false
+if get_option('with_wayland') != 'no'
+  wayland_dep = dependency(
+    'wayland-client',
+    version : '>= 1.11.0',
+    required : get_option('with_wayland') == 'yes')
+
```
uartie: Need to check for libva-wayland... otherwise compilation fails if libva-wayland is missing.  Better to fail during configure.

meson_options.txt
```diff
@@ -0,0 +1,5 @@
+option('driverdir', type : 'string', description : 'drivers path')
+option('disable_x11', type : 'boolean', value : false)
+option('with_wayland', type : 'combo', choices : ['yes', 'no', 'auto'], value : 'auto')
```
uartie: Can you make x11 and wayland options consistent.  For example, `with_x11` and `with_wayland` and have same combo options.

test/meson.build
```diff
@@ -0,0 +1,86 @@
...
+  'test_i965_drv_video',
+  [ test_i965_headers, test_i965_sources ],
+  dependencies : [ thread_dep, libdrm_dep, libva_drm_dep ],
+  link_with : [ libgtest, libi965_drv_video ],
+  include_directories : gtest_includes,
+  cpp_args : [ test_cppflags, libgtest_cppflags ],
+  override_options : [ 'cpp_std=c++11' ])
+
+# 1 minute for timeout
+test('driver test', test_i965_drv_video, timeout : 120)
```
uartie: It can take 2-3 minutes to run these tests on some platforms.

uartie:
LGTM... thanks Víctor!

Issue #262, opened 2017-08-31
VP9 Decoding Looks Awful


Queuecumber:
No idea whats causing this or if it's something fixable but I am trying to use VAAPI to decode VP9 video and it looks really bad. I grabbed the test video from [here](https://base-n.de/webm/out9.webm) and did two tests in Chromium and in `mpv` both with VAAPI enabled. 

In either case I get an image similar to

![screenshot from 2017-08-31 10-13-24](https://user-images.githubusercontent.com/5865387/29927673-0bf5e7ba-8e35-11e7-960a-91fcfaaf25d5.png)

This particular image was produced with `mpv --hwdec=vaapi Downloads/out9.webm`

For reference image should look something like 

![screenshot from 2017-08-31 10-14-08](https://user-images.githubusercontent.com/5865387/29927704-2384c5c2-8e35-11e7-951a-4127849faf23.png)

Is there any way to fix this either through configuration or code? Apologies in advance it this was already reported. 

My processor model is listed as 

> Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz

in /proc/cpuinfo  and I am running Ubuntu 17.04.

pengche1:
@Queuecumber , it seems it is the similar issue I am working on. could you tell me what version is your Linux kernel? I also have this issue on 4.13-rc7, but it just disappears on 4.12-X.
Also, please lspci | grep VGA and add it in the comment.

pengche1:
Could you help me to try this patch #263? thanks!

Queuecumber:
Yes that seems to be the problem, I was on 4.12-rc(something) when I noticed the issue and assumed I just hadn't played a VP9 video until then. I just booted into 4.12.7 and the same video plays fine with vaapi. So I guess we're calling it a 4.13 regression then. I'd love to test the patch, is this a kernel patch or something else?

Queuecumber:
Oh and for completeness sake, `lspci | grep VGA` gives

```
00:02.0 VGA compatible controller: Intel Corporation Device 591b (rev 04)
11:00.0 VGA compatible controller: NVIDIA Corporation GP104 [GeForce GTX 1080] (rev a1)
```

but ignore the nvidia card I don't use it for the desktop environment

pengche1:
#263 is a intel-vaapi-driver patch. it is a potential vaapi driver issue. And it just appears when migrate to 4-13-xx Linux kernel.

xhaihao:
[...]

PR #296, opened 2017-11-13
Add upper bound check in HCP_IND_OBJ_BASE_ADDR_STATE

xhaihao:
The commit 29e375a fixed some VP9 decoding artifacts reported in
https://github.com/01org/intel-vaapi-driver/issues/262, but we still
see artifacts when decoding some VP9 clips. More investigations show
that the artifacts are caused by graphics address 0

The graphics address for a GEM BO may be 0 since the commit
below was merged into i915 KMD

commit 2889caa9232109afc8881f29a2205abeb5709d0c
Author: Chris Wilson <chris@chris-wilson.co.uk>
Date:   Fri Jun 16 15:05:19 2017 +0100

    drm/i915: Eliminate lots of iterations over the execobjects array

So we never saw this issue before commit 2889caa

Setting upper bound to 0 causes that HW ignores upper bound check, which
works for non-zero base graphics address in HCP pipeline, however the
upper bound check is required for graphics address 0, so we add the bound
check no matter what is the graphics address

### DIFF
--- a/src/gen9_mfd.c
+++ b/src/gen9_mfd.c
@@ -341,7 +341,7 @@ gen9_hcpd_ind_obj_base_addr_state(VADriverContextP ctx,
 
     OUT_BCS_BATCH(batch, HCP_IND_OBJ_BASE_ADDR_STATE | (14 - 2));
     OUT_BUFFER_MA_REFERENCE(slice_data_bo);        /* DW 1..3 */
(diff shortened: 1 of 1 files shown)


xhaihao:
@pengche1 @QuPengfei could you help to review this PR? This PR should fix https://github.com/01org/intel-vaapi-driver/issues/297 as well.

Issue #284, opened 2017-10-17
VP9 CBR video corruption


asquared:
Garbled output is produced near a dissolve from a fullscreen graphic when encoding to VP9 on Kabylake at constant bitrate using FFmpeg. I attempted to reproduce this using vp9enc from vaapi-utils, but was unable to do so. Also I'm not able to reproduce using any other vaapi codec.

Here's the input (h.264) file, resulting VP9 output (garbled around 8 seconds in) and the ffmpeg log:
 [vp9bug.tar.gz](https://github.com/01org/intel-vaapi-driver/files/1392780/vp9bug.tar.gz)

Possibly related, when encoding the same file in VBR, the output is at approximately half the bitrate requested.

xhaihao:
@lizhong1008 It is most likely an issue in FFmpeg,  could you help to look into this issue?

asquared:
I had already opened https://trac.ffmpeg.org/ticket/6755 on the ffmpeg side.

lizhong1008:
Looks like a driver issue.  I tested with different bit rate(8800k, CBR)
./ffmpeg -vaapi_device /dev/dri/renderD128 -i /root/vp9bug/test.mp4 -vf "format=nv12,hwupload,deinterlace_vaapi,scale_vaapi=w=1920:h=1080" -c:v vp9_vaapi -b:v 8800k -bufsize 17600k -maxrate 8800k -g 150 -keyint_min 150 -bf 1 -bsf:v vp9_raw_reorder,vp9_superframe -y -an -f webm -dash 1 -r 30 vp9_v8800_CBR.webm

 and tested with another clips with same bitrate (4400k, CBR):
./ffmpeg -vaapi_device /dev/dri/renderD128 -i /root/Videos/skyfall2-trailer.mp4 -vf "format=nv12,hwupload,deinterlace_vaapi,scale_vaapi=w=1920:h=1080" -c:v vp9_vaapi -b:v 4400k -bufsize 17600k -maxrate 4400k -g 150 -keyint_min 150 -bf 1 -bsf:v vp9_raw_reorder,vp9_superframe -an -f webm -dash 1 -r 29.97 sky__v4400_CBR.webm

There are no garbled output.

asquared:
I can reproduce the garbled output at 8800k CBR if I remove "-r 30". I can produce correct output at 4400k CBR if I change "-r 29.97" to "-r 30". I wonder if this is related to #87.

[...]

PR #298, opened 2017-11-16
vp9enc: limit min_qp of brc to be non-zero

lizhong1008:
If base_qindex (luma_ac_qindex) is equal to zero, the enode mode will be
lossless mode when
luma_dc_qindex_delta/chroma_ac_qindex_delta/chroma_dc_qindex_delta are
also zero. In this case, the tx_mode should be set to ONLY_4X4, but current
driver only support TX_MODE_SELECT. This will cause the lossless mode
encoding failure.
Limitting the min_qp to be non-zero can force the encoding mode to be normal
mode.
It is to fix issue #284

Signed-off-by: Zhong Li <zhong.li@intel.com>


xhaihao:
it doesn't make sense to me using both 'cbr + lossless mode' at the same time.  It would be better an application can set valid parameters to driver.

lizhong1008:
@xhaihao here the lossless mode is produced by driver (BRC may outpu a base_qindex equal to zero) instead of application. This patch is to prevent lossless mode produced.

PR #302, opened 2017-11-21
configure: set strict default compiler flags

### LINKED ISSUE #301: Consider adding stricter compiler flags
Consider adding `-Wall -Werror` as default compiler flags in project.  This will encourage better coding practices.

uartie:
Use -Wall -Werror compiler flags for stricter compilation.
Also ensure per-target C/CXXFLAGS include AM_C/CXXFLAGS.

Fixes #301

Signed-off-by: U. Artie Eoff <ullysses.a.eoff@intel.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -212,6 +212,9 @@ if test "$enable_wayland" = "yes"; then
 fi
 AM_CONDITIONAL(USE_WAYLAND, test "$USE_WAYLAND" = "yes")
 
+AC_SUBST([AM_CFLAGS], ["-Wall -Werror"])
+AC_SUBST([AM_CXXFLAGS], ["-Wall -Werror"])
+
 AC_OUTPUT([
     Makefile
     src/Makefile
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -52,7 +52,7 @@ include Makefile.sources
 
 # convenience library that can be linked by driver and tests
 noinst_LTLIBRARIES		= libi965_drv_video.la
-libi965_drv_video_la_CFLAGS	= $(driver_cflags)
+libi965_drv_video_la_CFLAGS	= $(driver_cflags) $(AM_CFLAGS)
 libi965_drv_video_la_LDFLAGS	= $(driver_ldflags)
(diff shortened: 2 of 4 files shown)


uartie:
Need to fix compilation error triggered by -Wall -Werror strictness:
```
19:58:07,751 INFO  - gen8_mfc.c: In function 'gen8_mfc_jpeg_huff_table_state':
19:58:07,751 INFO  - gen8_mfc.c:3007:5: error: 'memset' used with length equal to number of elements without multiplication by element size [-Werror=memset-elt-size]
19:58:07,751 INFO  - memset(dc_table, 0, 12);
19:58:07,751 INFO  - ^~~~~~
19:58:07,751 INFO  - gen8_mfc.c:3008:5: error: 'memset' used with length equal to number of elements without multiplication by element size [-Werror=memset-elt-size]
19:58:07,751 INFO  - memset(ac_table, 0, 162);
19:58:07,751 INFO  - ^~~~~~
```

uartie:
compilation errors fixed.

fhvwy:
I'm in favour for developer builds, but I don't think -Werror should be enabled by default in release packages.  It causes future problems requiring additional patches against old releases when people build with new compilers which contain more warnings.

xhaihao:
@fhvwy You are right. E.g.  since '-Wall -Werror' was added in libva, there is a compiler error below when using an older version of GCC, it is most possible that we will have lots of similar issues when using a newer version of GCC in the future.

> va_fglrx.c: In function 'VA_FGLRXGetClientDriverName':                                                                       
> va_fglrx.c:243:33: error: 'ADL_Main_Control_Destroy' may be used uninitialized in this function [-Werror=maybe-uninitialized]
>          ADL_Main_Control_Destroy();                                                                                         
>                                  ^

uartie:
We could just add this to user-defined C/CXXFLAGS in CI builds instead of configure.ac since we have direct control over CI compiler versions.  @fhvwy @xhaihao what do you think?

uartie:
I will submit the memset fix separately

Issue #307, opened 2017-11-30
Driver aborts when encoding H.264 at level 6 or higher


fhvwy:
Setting level_idc to 60 (for level 6) gives:
```
ffmpeg_g: i965_avc_encoder_common.c:226: i965_avc_get_max_mv_len: Assertion `0' failed.
Aborted
```

fritsch:
Level 6, 6.1 and 6.2 specify up to 8k surfaces. Currently all drivers are limited to a maximum of 4096 - does that need changing in above combination?

fhvwy:
@fritsch  The surface size constraint is already handled elsewhere - i965_CreateContext() blocks creating contexts for too-large streams.

(In any case, the 4096 limit is already insufficient for full support of level 4.2, which requires being able to handle a stream with width or height of up to 16\*floor(sqrt(8702\*8)) = 4208 pixels, and higher levels even bigger.  Only levels up to 4.1 are fully supported.)

PR #308, opened 2017-11-30
Improve H.264 level-handling code

fhvwy:
Adds support for levels 6, 6.1 and 6.2, fixing #307.


src/i965_avc_encoder_common.c
```diff
@@ -153,108 +135,40 @@ i965_avc_calculate_initial_qp(struct avc_param * param)
...
-    case INTEL_AVC_LEVEL_4:
-    case INTEL_AVC_LEVEL_41:
-    case INTEL_AVC_LEVEL_42:
-    case INTEL_AVC_LEVEL_5:
-    case INTEL_AVC_LEVEL_51:
-    case INTEL_AVC_LEVEL_52:
-        max_v_mv_r = 512 * 4;
-        break;
-    default:
-        assert(0);
```
lizhong1008: How about print a warning message since the assertion has been removed?

fhvwy: I think that warning would be better placed further out - probably somewhere in the H.264 encoder setup, maybe gen9_avc_update_parameters()?  I'll refactor to add an "h264_level_is_valid" function and use it there.

fhvwy:
Updated: rebased, code rearranged a bit to be simpler, warning added for invalid level_idc.

src/i965_avc_encoder_common.c
lizhong1008: Yes, would better don't place the warning here. But I think put it in  intel_encoder_check_avc_parameter should be better.  BTW, do we need to reset it to be a default level such as 2 or just let the invalid level_idc written to the bitream? level 2 is what is using for  i965_avc_get_max_mbps and i965_avc_get_max_v_mv_r.
As a summary of my point, we can have two choices:
1. Print a warning message and reset it to be level 2. 
2. Just return an error code of  VA_STATUS_ERROR_INVALID_PARAMETER in intel_encoder_check_avc_parameter.

fhvwy: I don't think we want to do either of those.  The level_idc value is already written to the bitstream in the packed header, so it can't be changed.  It's also not necessarily invalid, even if it's not recognised - it could be a new version of the standard or an extension (consider the current problem, where a newer version of the standard defines new levels and the old driver aborts when it sees them).  I think just a simple warning and guessing some sensible limit values is the best approach.

lizhong1008: Got it. At least one thing needs to be changed, gen9_avc_update_parameters is not the best place to     give the warning. It means similar thing need to be done for gen6/7/8/x.

fhvwy: Currently only gen9 performs this test - gen<9 is not affected at all by the change.  If that is wanted, it could be done separately.

xhaihao: now gen8 and gen9 share the same code for AVC encoding, so this test is for both gen8 and gen9. BTW i965_avc_get_max_v_mv_r() is not called at all, how about to remove this function at all?

fhvwy:
I've added the level check to the generic code (also logging on some related checks around it).  The other unused function is removed, too.

lizhong1008:
LGTM

PR #309, opened 2017-11-30
Logging through libva

fhvwy:
This is starting to use the error and info callbacks added recently to libva.

### DIFF
--- a/src/i965_drv_video.c
+++ b/src/i965_drv_video.c
@@ -6559,8 +6559,11 @@ i965_ExportSurfaceHandle(VADriverContextP ctx, VASurfaceID surface_id,
     if (!obj_surface || !obj_surface->bo)
         return VA_STATUS_ERROR_INVALID_SURFACE;
 
-    if (mem_type != VA_SURFACE_ATTRIB_MEM_TYPE_DRM_PRIME_2)
+    if (mem_type != VA_SURFACE_ATTRIB_MEM_TYPE_DRM_PRIME_2) {
+        i965_log_info(ctx, "vaExportSurfaceHandle: memory type %08x "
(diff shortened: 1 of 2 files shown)


src/i965_drv_video.c
```diff
@@ -6904,6 +6904,53 @@ VAStatus i965_QueryVideoProcPipelineCaps(
...
+    va_start(vl, format);
+
+    if (!ctx->error_callback) {
+        // No error callback: this is a error message which should be
+        // user-visible, so print it to stderr instead.
+        vfprintf(stderr, format, vl);
+    } else {
+        // Put the message on the stack.  If it overruns the size of the
+        // then it will just be truncated - callers shouldn't be sending
+        // messages which are too long.
```
xhaihao: I agree the caller shouldn't sent a too long message but I think it is better to handle this case in i965_log_error()/i965_log_info():
tmp[1023]=0;
ret = vsnprintf(tmp, sizeof(tmp) - 1, format, vl);

fhvwy: I'm not sure I quite understand what you mean here.  It already uses the truncated result?

xhaihao: It is possible that tmp doesn't include the terminating null bytes and we should avoid this case. I think the simplest way is to limit the maximum number of bytes written to 'sizeof(tmp)-1' and set the last byte in tmp to null.

fhvwy: vsnprintf() necessarily writes a terminating null byte - see C11 sections 7.21.6.12 and 7.21.6.5.

xhaihao: Sorry, I made a mistake when I checked the code and Linux Programmer's Manual. Yes, you are right.

PR #313, opened 2017-12-19
Add the initial support of CNL

pengche1:


src/gen10_hevc_enc_common.c
```diff
@@ -0,0 +1,1398 @@
...
+    int i = 0;
+
+    for (i = 0; i < 4; i++) {
+        idx_offset = 0;
+        switch (i) {
+        case 0:
+            packed_type = VAEncPackedHeaderHEVC_VPS;
+            break;
+        case 1:
+            packed_type = VAEncPackedHeaderHEVC_VPS;
```
fhvwy: VPS is repeated; perhaps SPS?

src/gen10_hevc_enc_common.c
```diff
@@ -0,0 +1,1398 @@
...
+    case 186:
+        *max_bps = 4278190080;
+        *max_bps_per_pic = 35651584;
+        break;
+    default:
+        *max_bps = 16588800;
+        *max_bps_per_pic = 552760;
+        break;
+    }
+}
```
fhvwy: Can this be moved to common code?  The bitrate level limits are already present in gen9_hevc_enc_utils.c.

src/gen10_hevc_enc_common.c
```diff
@@ -0,0 +1,1398 @@
...
+    int i = 0, slice_idx = 0, start_index = 0;
+
+    for (i = 0; i < 4; i++) {
+        idx_offset = 0;
+        switch (i) {
+        case 0:
+            packed_type = VAEncPackedHeaderHEVC_VPS;
+            break;
+        case 1:
+            packed_type = VAEncPackedHeaderHEVC_VPS;
```
fhvwy: SPS?

src/i965_encoder.c
```diff
@@ -1494,7 +1495,8 @@ intel_enc_hw_context_init(VADriverContextP ctx,
 
     case VAProfileH264StereoHigh:
     case VAProfileH264MultiviewHigh:
-        if (IS_GEN9(i965->intel.device_info)) {
+        if (IS_GEN9(i965->intel.device_info) ||
+            IS_GEN9(i965->intel.device_info)) {
```
fhvwy: IS_GEN9 repeated.

src/gen10_hevc_enc_common.c
pengche1: yes, here VPS means SPS. but I'd like to keep it for SPS should be inserted just after VPS and they has the same header type VAEncPackedHeaderSequence. only idx_offset is 1, this header means SPS

src/i965_encoder.c
pengche1: I will update this error, thanks!

src/gen10_hevc_enc_common.c
pengche1: Definitely, it will be better to combine them together. but GEN10 needs max bit per picture which is not required by GEN9. I will check it more to decide how to deal it better.

src/gen10_hevc_enc_common.c
pengche1: @fhvwy , thanks for all your comments.  Now, we don't have plan to release source about those shader codes.

src/gen9_mfd.c
```diff
@@ -349,6 +353,66 @@ gen9_hcpd_ind_obj_base_addr_state(VADriverContextP ctx,
...
+    OUT_BCS_BATCH(batch, HCP_IND_OBJ_BASE_ADDR_STATE | (29 - 2));
+
+    /* DW1..5 indirect bitstream*/
+    OUT_RELOC64(batch,
+                slice_data_bo,
+                I915_GEM_DOMAIN_INSTRUCTION, 0,
+                0);
+    OUT_BCS_BATCH(batch, 0);
+    OUT_BCS_BATCH(batch, 0);
+    OUT_BCS_BATCH(batch, 0);
```
xhaihao: Please add the upper bound check, like as what we did for GEN9.

src/i965_device_info.c
```diff
@@ -640,6 +640,68 @@ static struct hw_codec_info cfl_hw_codec_info = {
...
+    .has_jpeg_decoding = 1,
+    .has_jpeg_encoding = 1,
+    .has_vpp = 1,
+    .has_accelerated_getimage = 1,
+    .has_accelerated_putimage = 1,
+    .has_tiled_surface = 1,
+    .has_di_motion_adptive = 1,
+    .has_di_motion_compensated = 1,
+    .has_vp8_decoding = 1,
+    .has_vp8_encoding = 1,
```
xhaihao: vp8 encoding isn't enabled by this commit, please reset this flag to 0 and reset it to 1 after vp8 encoding is added

src/gen9_mfd.c
pengche1: ok, I will add it and remove the unneeded patch

src/i965_device_info.c
pengche1: ok, I will update it

src/Makefile.sources
```diff
@@ -138,6 +138,7 @@ source_h = \
 	gen9_hevc_enc_utils.h \
 	gen9_hevc_encoder.h \
 	gen10_hcp_common.h \
+	i965_vdenc_defines.h     \
```
xhaihao: The new structures defined in i965_vdenc_defines.h isn't used for VDEnc H264, how about to remove this file, we can add the new structures back if needed in the future.

src/intel_driver.h
```diff
@@ -241,4 +242,6 @@ struct intel_region {
 
 #define IS_GEN10(device_info)           (device_info->gen == 10)
 
+#define IS_CNL(device_info)             (device_info->is_cannonlake)
```
xhaihao: IS_GEN10() is used in the previous commits for CNL, how about to remove IS_CNL() and use IS_GEN10() instead?
we can remove is_cannonlake flag from struct intel_device_info if do so.

Issue #316, opened 2018-01-09
VC1 regression caused by PR https://github.com/01org/intel-vaapi-driver/pull/295


xhaihao:
We are experiencing GPU hang issue  when playing back some VC1 clips. The issue disappears if reverting https://github.com/01org/intel-vaapi-driver/pull/295.

PR #317, opened 2018-01-09
VC1: Fix for frame coding mode

xhaihao:
Some VC1 clips might be marked interlaced but have fcm < 2, and the
wrong fcm will lead to undefined behavior, such as GPU hang.

This fixes https://github.com/01org/intel-vaapi-driver/issues/316


carpalis:
My mistake on the fcm logic. The fix looks correct to me. Thanks!

PR #329, opened 2018-01-16
Use generic register instead of acc0 to calculate the Y/U/V or R/G/B offset

xhaihao:
The instruction 'mov' cannot assign acc0 with :v anymore on CNL, so all AVS shaders
doesn't work as expected. Replacing acc0 with GRF registers works well on CNL and it
works on GEN8/GEN9 too.


src/shaders/post_processing/gen8/Save_AVS_RGBX.g8a
```diff
@@ -392,9 +392,9 @@
     //Internal LAYOUT:(RRGGBBAA)
         //Assign buffer channel order for Buffer 0123 in the order RGBA a0.3>A, a0.2>B, a0.1>G, a0.0>R
         // R = 0, G= 4, B = 8, A = 12.
-        mov (4) acc0.0<1>:w                 0x62EA:v
-        add (4) acc0.0<1>:w                 acc0<4;4,1>:w       70:uw
-        shl (4) r22.0<1>:w       acc0<4;4,1>:w       5:uw
+        mov (4) r22.0<1>:w                 0x62EA:v
+        add (4) r22.0<1>:w                 r22<4;4,1>:w       70:uw
```
yakuizhao: it will be better that the r22.0<4;4,1> is used.

xhaihao:
Update code style

PR #330, opened 2018-01-16
ENC: add mininum resolution query for encoder

QuPengfei:

### DIFF
--- a/src/i965_drv_video.c
+++ b/src/i965_drv_video.c
@@ -6408,6 +6408,25 @@ i965_QuerySurfaceAttributes(VADriverContextP ctx,
     attribs[i].value.value.i = max_height;
     i++;
 
+    if (obj_config->entrypoint == VAEntrypointEncSlice ||
+        obj_config->entrypoint == VAEntrypointEncSliceLP) {
+        attribs[i].type = VASurfaceAttribMinWidth;
+        attribs[i].value.type = VAGenericValueTypeInteger;
+        attribs[i].flags = VA_SURFACE_ATTRIB_GETTABLE;
+        attribs[i].value.value.i = 32;
+        if (obj_config->profile == VAProfileJPEGBaseline)
+            attribs[i].value.value.i = 16;
+        i++;
+
+        attribs[i].type = VASurfaceAttribMinHeight;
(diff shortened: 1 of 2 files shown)


QuPengfei:
update to do sanity check in the driver when middle ware does not do the query. this will avoid cause the GPU hang.

src/i965_drv_video.c
```diff
@@ -6408,6 +6408,25 @@ i965_QuerySurfaceAttributes(VADriverContextP ctx,
     attribs[i].value.value.i = max_height;
     i++;
 
+    if (obj_config->entrypoint == VAEntrypointEncSlice ||
+        obj_config->entrypoint == VAEntrypointEncSliceLP) {
+        attribs[i].type = VASurfaceAttribMinWidth;
+        attribs[i].value.type = VAGenericValueTypeInteger;
+        attribs[i].flags = VA_SURFACE_ATTRIB_GETTABLE;
+        attribs[i].value.value.i = 32;
+        if (obj_config->profile == VAProfileJPEGBaseline)
```
uartie: VAProfileJPEGBaseline only has VAEntrypointEncPicture encode entrypoint.  Thus, this will never be true.

src/i965_encoder.c
```diff
@@ -824,6 +824,16 @@ intel_encoder_check_misc_parameter(VADriverContextP ctx,
...
     VAStatus ret = VA_STATUS_SUCCESS;
+    int min_width_height = 32;
+
+    if (encoder_context->frame_width_in_pixel > 0 &&
+        encoder_context->frame_height_in_pixel > 0) {
+        if (profile == VAProfileJPEGBaseline)
+            min_width_height = 16;
+        if (encoder_context->frame_width_in_pixel < min_width_height ||
+            encoder_context->frame_height_in_pixel < min_width_height)
+            return VA_STATUS_ERROR_INVALID_PARAMETER;
```
uartie: Should this return VA_STATUS_ERROR_RESOLUTION_NOT_SUPPORTED?

uartie: Also, this check should not be needed if it's properly done in create context.

PR #338, opened 2018-01-19
Vp9 enc on cnl

xhaihao:
This patch series adds support for vp9 encoding on CNL


src/gen10_vdenc_common.c
```diff
@@ -0,0 +1,215 @@
...
+#include <string.h>
+#include <math.h>
+#include <assert.h>
+
+#include "intel_batchbuffer.h"
+#include "i965_defines.h"
+#include "i965_drv_video.h"
+#include "gen10_vdenc_common.h"
+
+#define VDENC_WRITE_COMMANDS(command_flag)                        \
```
QuPengfei: it is good add *param and batch as the Macro define parameter
define VDENC_WRITE_COMMANDS(command_flag,param,batch)

src/gen10_vdenc_vp9.h
```diff
@@ -0,0 +1,1644 @@
...
+#define VDENC_VDENC_CMD0_STATE_SIZE_IN_DWS              (VDENC_VDENC_CMD0_STATE_SIZE >> 2)
+#define VDENC_VDENC_CMD1_STATE_SIZE_IN_DWS              (VDENC_VDENC_CMD1_STATE_SIZE >> 2)
+
+enum VDENC_VP9_LUT_MODE {
+    VDENC_VP9_LUT_MODE_INTRA_SADMPM             = 0x00,
+    VDENC_VP9_LUT_MODE_INTRA_32x32              = 0x01,
+    VDENC_VP9_LUT_MODE_INTRA_16x16              = 0x02,
+    VDENC_VP9_LUT_MODE_INTRA_8x8                = 0x03,
+
+    VDENC_VP9_LUT_MODE_INTER_32x16              = 0x04,
```
QuPengfei: these three emum value use the same 0x04. Is it reasonable?

xhaihao: This enum isn't used any more, I will remove it from this commit.

src/gen10_vdenc_common.c
xhaihao: Thanks for the comment, I will update the commit to redefine this Macro.

xhaihao:
@QuPengfei thanks for your review, I Updated the corresponding commits per your comments.

src/i965_defines.h
```diff
@@ -911,6 +911,7 @@
 #define HEVC_SLICE_I                            2
 
 #define HCP_CODEC_HEVC                          0
+#define HCP_CODEC_VP9                           1
```
QuPengfei: the   HCP_CODEC_VP9 is defined twice

src/gen10_vdenc_vp9.c
```diff
@@ -0,0 +1,4629 @@
...
+                                   struct gen10_vdenc_vp9_dys_curbe_parameters *curbe_param)
+{
+    struct gen10_vdenc_vp9_dys_curbe_data *curbe_cmd;
+
+    curbe_cmd = i965_gpe_context_map_curbe(gpe_context);
+
+    if (!curbe_cmd)
+        return;
+
+    memset(cur
```
QuPengfei: it is  good remove the commented line if it is not needed

xhaihao:
Update, thx @QuPengfei

Issue #333, opened 2018-01-18
Unit test regressions since #330 on SKL+


uartie:
There are at least 30 unit test regressions on SKL+ (test_i965_drv_video) since PR #330 merged:

```
[==========] 622 tests from 16 test cases ran. (53513 ms total)
[  PASSED  ] 592 tests.
[  FAILED  ] 30 tests, listed below:
[  FAILED  ] Edge/JPEGEncodeInputTest.Full/280, where GetParam() = (Fixed Size 1x1, 0x5ff61d pointing to "I420")
[  FAILED  ] Edge/JPEGEncodeInputTest.Full/281, where GetParam() = (Fixed Size 1x1, 0x5ff618 pointing to "NV12")
[  FAILED  ] Edge/JPEGEncodeInputTest.Full/282, where GetParam() = (Fixed Size 1x1, 0x5ff613 pointing to "UYVY")
[  FAILED  ] Edge/JPEGEncodeInputTest.Full/283, where GetParam() = (Fixed Size 1x1, 0x5ff60e pointing to "YUY2")
[  FAILED  ] Edge/JPEGEncodeInputTest.Full/284, where GetParam() = (Fixed Size 1x1, 0x5ff609 pointing to "Y800")
[  FAILED  ] Edge/JPEGEncodeInputTest.Full/285, where GetParam() = (Fixed Size 1x2, 0x5ff61d pointing to "I420")
[  FAILED  ] Edge/JPEGEncodeInputTest.Full/286, where GetParam() = (Fixed Size 1x2, 0x5ff618 pointing to "NV12")
[  FAILED  ] Edge/JPEGEncodeInputTest.Full/287, where GetParam() = (Fixed Size 1x2, 0x5ff613 pointing to "UYVY")
[  FAILED  ] Edge/JPEGEncodeInputTest.Full/288, where GetParam() = (Fixed Size 1x2, 0x5ff60e pointing to "YUY2")
[  FAILED  ] Edge/JPEGEncodeInputTest.Full/289, where GetParam() = (Fixed Size 1x2, 0x5ff609 pointing to "Y800")
[  FAILED  ] Edge/JPEGEncodeInputTest.Full/290, where GetParam() = (Fixed Size 2x1, 0x5ff61d pointing to "I420")
[  FAILED  ] Edge/JPEGEncodeInputTest.Full/291, where GetParam() = (Fixed Size 2x1, 0x5ff618 pointing to "NV12")
[...]

PR #340, opened 2018-01-19
ENC:fix  the regression by min resolution check for encoder

QuPengfei:
JPEG encoder allow min resolution 1x1
add min resolution check in vaCreateContext to avoid gtest failture
add fix for gtest case
fix the regression in 88249259d4abb2fd701ed3db229941738da2497e

Fixes #330 #333

### DIFF
--- a/src/i965_drv_video.c
+++ b/src/i965_drv_video.c
@@ -2479,11 +2479,15 @@ i965_CreateContext(VADriverContextP ctx,
     int i;
     int max_width;
     int max_height;
+    int min_width_height = I965_MIN_CODEC_ENC_RESOLUTION_WIDTH_HEIGHT;
 
     if (NULL == obj_config) {
         vaStatus = VA_STATUS_ERROR_INVALID_CONFIG;
         return vaStatus;
     }
+
+    if (obj_config->profile == VAProfileJPEGBaseline)
+        min_width_height = 1;
 
     max_resolution(i965, obj_config, &max_width, &max_height);
 
@@ -2491,6 +2495,17 @@ i965_CreateContext(VADriverContextP ctx,
         picture_height > max_height) {
         vaStatus = VA_STATUS_ERROR_RESOLUTION_NOT_SUPPORTED;
         return vaStatus;
+    }
+
+    if ((VAEntrypointEncSlice == obj_config->entrypoint) ||
+        (VAEntrypointEncPicture == obj_config->entrypoint) ||
+        (VAEntrypointEncSliceLP == obj_config->entrypoint) ||
+        (VAEntrypointFEI == obj_config->entrypoint)) {
+        if (picture_width < min_width_height ||
+            picture_height < min_width_height) {
+            vaStatus = VA_STATUS_ERROR_RESOLUTION_NOT_SUPPORTED;
+            return vaStatus;
+        }
     }
 
     /* Validate flag */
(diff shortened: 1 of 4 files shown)


src/i965_drv_video.c
```diff
@@ -6409,21 +6425,22 @@ i965_QuerySurfaceAttributes(VADriverContextP ctx,
     i++;
 
     if (obj_config->entrypoint == VAEntrypointEncSlice ||
-        obj_config->entrypoint == VAEntrypointEncSliceLP) {
+        obj_config->entrypoint == VAEntrypointEncSliceLP ||
+        obj_config->entrypoint == VAEntrypointFEI) {
```
uartie: missing `obj_config->entrypoint == VAEntrypointEncPicture` (i.e. used by VAProfileJPEGBaseline)

src/i965_drv_video.c
```diff
@@ -6409,21 +6425,22 @@ i965_QuerySurfaceAttributes(VADriverContextP ctx,
...
         attribs[i].type = VASurfaceAttribMinWidth;
         attribs[i].value.type = VAGenericValueTypeInteger;
         attribs[i].flags = VA_SURFACE_ATTRIB_GETTABLE;
-        attribs[i].value.value.i = 32;
+        attribs[i].value.value.i = I965_MIN_CODEC_ENC_RESOLUTION_WIDTH;
         if (obj_config->profile == VAProfileJPEGBaseline)
-            attribs[i].value.value.i = 16;
+            attribs[i].value.value.i = 1;
```
uartie: What was the original reason for using `16`?

src/i965_drv_video.h
```diff
@@ -81,6 +81,7 @@
 #define ENCODER_HIGH_QUALITY      ENCODER_DEFAULT_QUALITY
 #define ENCODER_LOW_QUALITY       2
 
+#define I965_MIN_CODEC_ENC_RESOLUTION_WIDTH          32
```
uartie: The name of this macro does not convey its usage.  That is, it's also used to constrain minimum *height* in your code.

test/i965_avce_context_test.cpp
```diff
@@ -115,7 +119,7 @@ TEST_P(AVCEContextTest, RateControl)
         config = createConfig(profile, entrypoint, attribs, expect);
         if (expect != VA_STATUS_SUCCESS) continue;
 
-        context = createContext(config, 1, 1);
+        context = createContext(config, min_resolution_width_height, min_resolution_width_height);
```
uartie: Please wrap this line and all similar ones below.

test/i965_avce_context_test.cpp
```diff
@@ -65,6 +65,9 @@ class AVCEContextTest
             ||IS_GEN10(i965->intel.device_info))
             is_gen9 = true;
 
+        if (VAProfileJPEGBaseline == profile)
```
uartie: This is unnecessary since all these test cases deal with H264 profiles only.

src/i965_drv_video.c
QuPengfei: it is the limitation in the closed source code. and there is no limitation here.

src/i965_drv_video.c
QuPengfei: thanks.

src/i965_drv_video.h
QuPengfei: thanks

test/i965_avce_context_test.cpp
QuPengfei: @uartie ,i am not sure what is the idea of wrap? would you explain it?

test/i965_avce_context_test.cpp
QuPengfei: thanks

test/i965_avce_context_test.cpp
QuPengfei: thanks

QuPengfei:
update the PR

Issue #342, opened 2018-01-21
vaExportSurfaceHandle possibly misinterprets surface size


pkerling:
Hi,

while investigating usage of the recently implemented `vaExportSurfaceHandle` in the Kodi media center we came across the problem that EGL import with `eglCreateImageKHR` of the surfaces decoded from HEVC10 failed with `EGL_BAD_ALLOC` in recent Mesa versions, while 8-bit HEVC and H.264 worked fine.

Testing the import of the Y plane of an example 4K (3840x2160) video decoded as P010, I traced the failure down to the following mesa code: https://github.com/mesa3d/mesa/blob/master/src/intel/isl/isl.c#L1325-L1335
It fails because the minimum pitch calculated is 15360, but the actual pitch is 7680 (which is 2*3840, i.e. the actual pitch is correct since the P010 Y plane has 2 bytes per pixel).

Now the wrong minimum pitch calculation seems to be a result of `intel-vaapi-driver` giving an erroneous result for the `width` and `height` fields of the `VADRMPRIMESurfaceDescriptor`, which I pass to `eglCreateImageKHR` as `EGL_WIDTH` and `EGL_HEIGHT`.
They are filled in here: https://github.com/01org/intel-vaapi-driver/blob/master/src/i965_drv_video.c#L6742-L6743
But `object_surface.width`is the *plane pitch in bytes* while it was my understanding that `VADRMPRIMESurfaceDescriptor.width` is the *surface width in pixels*. Now the struct description does not explicitly spell this out, but at least the mesa vaapi drivers use pixels here, so it would be cool to agree on something.

Using `orig_width`/`orig_height` instead of `width`/`height` solves the problem for me.

xhaihao:
Yes, VADRMPRIMESurfaceDescriptor.width should be the surface width in pixels.

fhvwy:
Aside:  I don't think you want to use the surface width/height for this sort of use-case - you want the size of the image actually rendered on the surface, which may be smaller (especially when there is a cropping rectangle present).  This is probably why the problem wasn't noticed earlier - other programs using the API so far do not use these values because they only need the size of the image rendered on the surface, not the size of surface itself.

pkerling:
Thanks for the hint - under what conditions would that cropping occur? And are there downsides to just importing the whole surface to EGL using the parameters provided in VADRMPRIMESurfaceDescriptor?

[...]

PR #343, opened 2018-01-22
Return surface width/height in pixels

xhaihao:
width/height in VADRMPRIMESurfaceDescriptor specifies the
surface width/height in pixels in VA-API, so we should set
the corresponding variables in pixels in the driver

This fixes https://github.com/01org/intel-vaapi-driver/issues/342,
thanks pkerling for the investigation and proposed fix for this issue.


fhvwy:
I'm not sure that these numbers are the right ones to use.  The surface is likely to be larger than `orig_width` by `orig_height` because of alignment rounding, which can be relevant when someone wanting to render to a surface works in units larger than individual pixels.  For example, the H.264 decoder needs to know that there is space for whole macroblocks around the edges - an external renderer may have a similar requirement, and therefore should be told the size of the whole surface so that they can deal with these cases properly.

I think that therefore wants `obj_surface->width / bpp_1stplane_by_fourcc(fourcc)` and `obj_surface->height`, though I haven't checked that that is right in all cases.

xhaihao:
if a larger surface is needed for a codec, the application should specify the larger size when creating a surface.  E.g. a 1920x1088 surface is created in mpv for 1080x1080 video decoding. 

On the other hand, obj_surface->width is the stride (or pitch) of the surface, so it should follow the alignment, e.g. for a tiled surface. so obj_surface->width / bpp_1stplane_by_fourcc(fourcc) is not what you want in this case.

fhvwy:
Hmm, right - adjustment for alignment is applied after the multiplication by element size so the simple division isn't helpful.

I guess I'm unclear what the intent of these fields is - having written code using the feature I have never used them.  You have to track the size of the image on the surface separately and will use that for things like EGL import, so I'm not really seeing any use-case here.  Given that, I suppose having the size as passed to vaCreateSurfaces() is as good as any other similar value we might offer.  Still, the one user example we have here demonstrated that their presence invites people to use it incorrectly - should something be done about that?

xhaihao:
@fhvwy Sorry, what's the example you mentioned in the above comment?

fhvwy:
@xhaihao  I mean the original report which prompted this (#342), which is the one example I have of someone who isn't me using this API.

PR #345, opened 2018-01-25
Fix enc regressions

xhaihao:
Recently we added checks for config attributes, which results in some regression for vp8 and jpeg encoding.  This patch series fixes these regressions:
1. No packed header is required for VP8 encoding
2. CQP is supported by JPEG encoding.


uartie:
I verified locally that ffmpeg and libyami tests pass now.  However, the bigger issue is that these middleware are hard-coding assumptions about supported driver attributes.  That is, if they actually queried the driver (vaGetConfigAttributes), then they probably would have not broke due to previous driver change.  I've already filed a bug with libyami.  Please file a bug with ffmpeg.  Gstreamer-vaapi appears to do the right thing.

PR #348, opened 2018-01-30
ENC/VP8: set value to VA_ATTRIB_NOT_SUPPORTED when querying VAConfigA…

xhaihao:
…ttribEncPackedHeaders

vaGetConfigAttributes() should return VA_ATTRIB_NOT_SUPPORTED if the
corresponding attribute is not supported for a given profile/entrypoint
pair. The driver doesn't support any packed header from application for
VP8 encoding.


xhaihao:
No extra work is required for FFmpeg/libav, gst-vaapi, libyami etc.

PR #353, opened 2018-02-02
Autotools clean up

ceyusa:
Some clean ups for autotools scritps.  It makes faster the configure script execution and enables the distcheck target out-of-tree


configure.ac
```diff
@@ -135,29 +122,14 @@ VA_MINOR_VERSION=`echo "$VA_VERSION" | cut -d'.' -f2`
...
-    PKG_CHECK_MODULES([X11],    [x11],    [:], [USE_X11="no"])
-    PKG_CHECK_MODULES([XEXT],   [xext],   [:], [USE_X11="no"])
-    PKG_CHECK_MODULES([XFIXES], [xfixes], [:], [USE_X11="no"])
-fi
-
 if test "$USE_X11" = "yes"; then
-    PKG_CHECK_MODULES(LIBVA_X11_DEPS, [libva-x11],
-      [AC_DEFINE([HAVE_VA_X11], [1], [Defined to 1 if VA/X11 API is enabled])],
-      [USE_X11="no"])
+   AC_DEFINE([HAVE_VA_X11], [1], [Defined to 1 if VA/X11 API is enabled])
```
uartie: We still need to check for libva-x11 availability.  If libva is compiled with `--disable-x11` the associated libva-x11 headers don't get installed.  Then, when user tries to `--enable-x11` on intel-vaapi-driver, the compilation will fail due to missing headers.   It's better to fail during configure if libva-x11 is not installed.

configure.ac
```diff
@@ -173,41 +145,19 @@ fi
...
-    PKG_CHECK_MODULES([LIBVA_WAYLAND_DEPS], [libva-wayland],
-        [USE_WAYLAND="yes"], [:])
-
-    if test "$USE_WAYLAND" = "yes"; then
-
-        PKG_CHECK_MODULES([WAYLAND_CLIENT], [wayland-client >= wayland_api_version],
-        [INCLUDES="$INCLUDES $WAYLAND_CLIENT_CFLAGS"], [:])
+   PKG_CHECK_MODULES([WAYLAND_CLIENT], [wayland-client >= wayland_api_version],
+     [
+        USE_WAYLAND="yes"
```
uartie: Same as libva-x11 comment... if libva is compiled with `--disable-wayland`, then driver will fail to compile if `--enable-wayland` is used during driver configure.

uartie:
LGTM

PR #358, opened 2018-02-13
Accept empty packed headers attribute in vaCreateConfig()

fhvwy:
This indicates to the driver that the user will not supply any packed
headers, even though the possibility is available.  This was an API/ABI
break in cccf2a3c61b3cdc4836ee6a0a30db436fcc6ffe5, and existing programs
such as libavcodec do use this pattern.

### DIFF
--- a/src/i965_drv_video.c
+++ b/src/i965_drv_video.c
@@ -1432,8 +1432,21 @@ i965_CreateConfig(VADriverContextP ctx,
         if (attrib_found) {
             uint32_t enc_packed_attribs = i965_get_enc_packed_attributes(ctx, profile, entrypoint);
 
-            if (!(attrib_found->value & enc_packed_attribs))
+            if (enc_packed_attribs == VA_ATTRIB_NOT_SUPPORTED) {
+                i965_log_info(ctx, "vaCreateConfig: invalid EncPackedHeaders attribute %#x: "
+                              "packed headers are not supported.\n", attrib_found->value);
                 vaStatus = VA_STATUS_ERROR_INVALID_VALUE;
+            } else if (attrib_found->value == 0) {
+                i965_log_info(ctx, "vaCreateConfig: setting the EncPackedHeaders attribute to zero to "
+                              "indicate that no packed headers will be used is deprecated.\n");
+            } else {
+                if (attrib_found->value & ~enc_packed_attribs) {
+                    i965_log_info(ctx, "vaCreateConfig: invalid EncPackedHeaders attribute %#x: "
+                                  "some packed headers are not supported (supported set %#x).\n",
+                                  attrib_found->value, enc_packed_attribs);
+                    vaStatus = VA_STATUS_ERROR_INVALID_VALUE;
+                }
+            }
         }
     }
 


uartie:
If the driver reports the attribute as unsupported, then shouldn't unsupported be the only acceptable value?

uartie:
See fe0cac201f09f4d4eb1935de1803a58fef9e61d2

uartie:
Also, libavcodec already does the right thing... see https://github.com/intel/libva/issues/178#issuecomment-365470500

fhvwy:
> If the driver reports the attribute as unsupported, then shouldn't unsupported be the only acceptable value?

Yes.  However, the problem here is for VP9 which does report some packed headers as supported.

> Also, libavcodec already does the right thing... see intel/libva#178 (comment)

I agree.  Unfortunately, the current Intel driver rejects this behaviour and therefore VP9 encoding is broken when using it on all versions of libavcodec.  (Hence considering it a ABI compatibility break, because programs which previously worked on the version 1.8 or 2.0 drivers no longer do.)

uartie:
Ah, I see, so VP9 RAW packed header is optional.  And therefore, all packed headers must be treated as optional since there is no way to explicitly relay it otherwise :-/

uartie:
Would you mind adding a unit test, at least for the VP9 case, so we can detect if this gets broken again in the future?

fhvwy:
@uartie  Complete test added as #361.  I made a separate PR because I don't really know what I'm doing with the test framework and I don't want it to block merging this fix.

(This branch also updated to pass the test.)

uartie:
@xhaihao bump

fhvwy:
@xhaihao  Updated with a warning for the empty set case.

src/i965_drv_video.c
```diff
@@ -1431,8 +1431,23 @@ i965_CreateConfig(VADriverContextP ctx,
...
             uint32_t enc_packed_attribs = i965_get_enc_packed_attributes(ctx, profile, entrypoint);
 
-            if (!(attrib_found->value & enc_packed_attribs))
-                vaStatus = VA_STATUS_ERROR_INVALID_VALUE;
+            if (enc_packed_attribs == VA_ATTRIB_NOT_SUPPORTED) {
+                if (attrib_found->value != VA_ATTRIB_NOT_SUPPORTED) {
+                    i965_log_info(ctx, "vaCreateConfig: invalid EncPackedHeaders attribute %#x: "
+                                  "packed headers are not supported.\n", attrib_found->value);
+                    vaStatus = VA_STATUS_ERROR_INVALID_VALUE;
+                }
```
xhaihao: If the driver doesn't support packed header at all, application should not pass VAConfigAttribEncPackedHeaders to the driver. so I think the driver should return error no matter what is attrib_found->value

fhvwy:
@xhaihao  Sure, updated.

xhaihao:
Could you update the commit log as well?  Per the commit log, vaCreateConfig() can accept VAConfigAttribEncPackedHeaders when packed header is not supported.

fhvwy:
@xhaihao  Sure, updated.

xhaihao:
Thanks for the patch, applied. @fhvwy

PR #361, opened 2018-02-15
Add test for config creation with packed headers

fhvwy:
Iterates across all profiles and entrypoints creating configs, ensuring
that all valid combinations of packed headers and accepted and all
invalid combinations are rejected.

@uartie  I don't really have any idea what I am doing in this test framework.  There appears to be a lot of magic to do clever stuff, but I can't make sense of it so I just have a single function which loops over everything doing the required checks directly.  If you'd like to take this and turn it into something more conforming to how you think the framework should be used then please do.

### DIFF
--- a/test/Makefile.am
+++ b/test/Makefile.am
@@ -72,6 +72,7 @@ test_i965_drv_video_SOURCES =						\
 	i965_test_environment.cpp					\
 	i965_test_fixture.cpp						\
 	i965_test_image_utils.cpp					\
+	i965_config_packed_header_test.cpp				\
 	object_heap_test.cpp						\
 	test_main.cpp							\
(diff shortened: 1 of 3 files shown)


uartie:
Yeh, there's a lot of changes that could be made here.  I'll try to take a stab at it next week if I have time.

uartie:
You can remove the patch here that is already part of the other PR.

fhvwy:
> You can remove the patch here that is already part of the other PR.

Sure, done.  (The test of course fails a large number of cases without the other patch.)

uartie:
understood... thanks.

uartie:
Superseded by intel/libva-utils#105

PR #369, opened 2018-03-09
[DRI3] Enable DRI3 and Present Extension support

jayesht:
XWayland was failing to function with DRI2 as by default gnome shell supports DRI3 only whereas libva only supported dri2.
https://github.com/intel/libva/issues/79

This pull request enables dri3 support for driver.

Dependency: https://github.com/intel/libva/pull/180 
Usage: --enable-dri3 flags is added in config to enable this feature.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -52,6 +52,11 @@ AC_ARG_ENABLE(x11,
                     [build with VA/X11 API support @<:@default=yes@:>@])],
     [], [enable_x11="yes"])
 
+AC_ARG_ENABLE(dri3,
+    [AC_HELP_STRING([--enable-dri3],
+                    [build with VA/X11-DRI3 API support @<:@default=no@:>@])],
+    [], [enable_dri3="no"])
+
 AC_ARG_ENABLE([wayland],
     [AC_HELP_STRING([--enable-wayland],
                     [build with VA/Wayland API support @<:@default=yes@:>@])],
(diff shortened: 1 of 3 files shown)


xhaihao:
DRI3 in the driver should be disabled if DRI3 is not enabled in libva, Otherwise you might get the error below:

error: failed to resolve va_dri3_createPixmap(): /usr/lib/x86_64-linux-gnu/libva-x11.so.2: undefined symbol: va_dri3_createPixmap

sjnewbury:
This works well with mpv using the gpu/gl output and hwdec=vaapi.  Using the vaapi output there seems there's an issue with vblank/flip timing resulting in a black band.  With VA-API enabled Chromium DRI3 results in no output and/or render corruption depending upon other Chromium flags.  It appears the either buffer is being written the wrong (or unexpected) address or the output buffer in the browser is pointing to the wrong place.

realh:
I tried applying this to Arch and it doesn't seem to have helped. I might have done something wrong, but before I look into that, should this fix work on its own or do I also need to patch libva?

rossburton:
Any progress on this?

sjnewbury:
@rossburton At least other testers have seen the same as I have, it's not just me!  @jayesht, any chance you can take a look at this, DRI3 support really should work, otherwise it's really getting in the way of Wayland adoption.  It's certainly preventing me from switching, and I'm sure I'm not the only one!

uartie:
Original submitter is no longer active and has abandoned this PR. Someone else will need to take ownership and resubmit in a new PR, please.

jnettlet:
I have reported over on https://github.com/intel/libva/issues/79 that I have rebased the patches and with the patches for Xwayland https://gitlab.freedesktop.org/xorg/xserver/issues/631 I know have functioning hardware decoding in Chromium-vaapi.  I am quite familiar with this patchset at this point, and will gladly resubmit a rebased pull request if there is nobody else to do it.

jayesht:
@jnettlet : Glad to know the patch set works. :) Unfortunately I do not have means to validate the changes so feel free to submit a rebased patchset.

paulmenzel:
@jnettlet, will you create a new merge/pull request or reuse this one?

jnettlet:
I am sorry,  been traveling and caught up in work projects.  I can push a
rebased patchset, but there are still some additional things that I think
need to be implemented.  It probably makes sense to get the base code
merged and then we can work on bug fixes and enhancements from there.

paulmenzel:
On 03/10/19 17:58, Jon Nettleton wrote:
> I can push a rebased patchset, but there are still some additional
> things that I think need to be implemented.  It probably makes sense
> to get the base code merged and then we can work on bug fixes and
> enhancements from there.
I totally agree, that this should be the way to go.

fixerivan:
what is the situation on this @jnettlet ? thx

PR #372, opened 2018-03-14
Fix a potential over run issue

pengche1:


src/gen9_hevc_encoder.c
```diff
@@ -6570,7 +6570,11 @@ gen9_hevc_set_reflist(VADriverContextP ctx,
 
     for (i = 0; i < 16; i++) {
         frame_idx = -1;
-        obj_surface = SURFACE(ref_list[i].picture_id);
+        if (i < 15)
```
QuPengfei: it seems missed the i==15 case

pengche1: No, for the last i = 15, it goes to "else" and don't get the ref_list[i].

QuPengfei: it is ok to me

PR #401, opened 2018-06-20
Fix the references for VP8 encoding

xhaihao:
If a reference is unavailable, use the next frame instead

### DIFF
--- a/src/i965_encoder_vp8.c
+++ b/src/i965_encoder_vp8.c
@@ -2087,6 +2087,19 @@ i965_encoder_vp8_get_picture_parameter(VADriverContextP ctx,
         vp8_context->ref_arf_frame = NULL;
     }
 
+    if (!is_intra) {
+        if (!vp8_context->ref_last_frame) {
+            if (vp8_context->ref_gf_frame)
+                vp8_context->ref_last_frame = vp8_context->ref_gf_frame;
+            else
+                vp8_context->ref_last_frame = vp8_context->ref_arf_frame;
+        }
+
+        if (!vp8_context->ref_gf_frame) {
+            vp8_context->ref_gf_frame = vp8_context->ref_arf_frame;
+        }
+    }
+
     vp8_context->brc_distortion_buffer_need_reset = 0;
 
     if (brc_enabled) {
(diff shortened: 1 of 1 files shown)


src/i965_encoder_vp8.c
```diff
@@ -2087,6 +2087,19 @@ i965_encoder_vp8_get_picture_parameter(VADriverContextP ctx,
         vp8_context->ref_arf_frame = NULL;
     }
 
+    if (!is_intra) {
+        if (!vp8_context->ref_last_frame) {
+            if (vp8_context->ref_gf_frame)
+                vp8_context->ref_last_frame = vp8_context->ref_gf_frame;
+            else
+                vp8_context->ref_last_frame = vp8_context->ref_arf_frame;
```
lizhong1008: Is it possible last/gf/arf frames are all NULL frame? Maybe we can add an assert() for this case.

src/i965_encoder_vp8.c
```diff
@@ -2087,6 +2087,19 @@ i965_encoder_vp8_get_picture_parameter(VADriverContextP ctx,
...
+        }
+
+        if (!vp8_context->ref_gf_frame) {
+            vp8_context->ref_gf_frame = vp8_context->ref_arf_frame;
```
lizhong1008: Should check ref_arf_frame is not NULL too?

src/i965_encoder_vp8.c
```diff
@@ -5946,14 +5959,26 @@ i965_encoder_vp8_pak_pipeline_prepare(VADriverContextP ctx,
...
+    }
+
+    if (vp8_context->ref_arf_frame) {
+        PAK_REFERENCE_BO(vp8_context->reference_surfaces[2].bo, vp8_context->ref_arf_frame->bo, 1);
+    } else {
+        PAK_REFERENCE_BO(vp8_context->reference_surfaces[2].bo, NULL, 0);
+    }
+
+    for (i = 3; i < ARRAY_ELEMS(vp8_context->reference_surfaces); i++) {
+        PAK_REFERENCE_BO(vp8_context->reference_surfaces[i].bo, NULL, 0);
```
lizhong1008: The reset LGTM

src/i965_encoder_vp8.c
xhaihao: For non-key frame, one of last/gf/arf must be non-NULL.

lizhong1008: This is normal case. Actually for normal case, last frame must be non-NULL for P frames.

uartie:
What is the status here?

xhaihao:
@uartie Some extra works are required, currently it is in low priority.

PR #429, opened 2018-12-18
Watchdog i965 media driver changes

cmsanta:
This CL adds the ioctl necessary to set the watchdog timer
after each batch buffer is submitted.

This is part of the Timeout Detection and Recovery (TDR)
mechanism previously done here in upstream:
https://patchwork.freedesktop.org/series/21868/.

Tested-by: Carlos Santa <carlos.santa@intel.com>
Signed-off-by: Carlos Santa <carlos.santa@intel.com>

### DIFF
--- a/src/i965_drv_video.c
+++ b/src/i965_drv_video.c
@@ -2744,13 +2744,17 @@ i965_CreateContext(VADriverContextP ctx,
     i965->current_context_id = contextID;
 
     i965->bufmgr = drm_intel_bufmgr_gem_init(intel->fd, 4096);
-
+    assert(i965->bufmgr);
     i965->gem_context = drm_intel_gem_context_create(i965->bufmgr);
+    assert(i965->gem_context);
     intel->gem_context = i965->gem_context;
     intel->gem_ctx_id = i965->gem_context->ctx_id;
 
     if (kernel_has_gpu_watchdog_support(&i965->intel))
         i965->intel.has_watchdog = 1;
+
+    if (i965->intel.has_watchdog)
+        set_watchdog_timer_threshold(&i965->intel, picture_width, picture_height);
 
     return vaStatus;
 }
@@ -2777,6 +2781,7 @@ i965_DestroyContext(VADriverContextP ctx, VAContextID context)
     }
 
     drm_intel_gem_context_destroy(i965->gem_context);
+    drm_intel_bufmgr_destroy(i965->bufmgr);
 
     i965->intel.has_watchdog = 0;
 
--- a/src/i965_drv_video.c
+++ b/src/i965_drv_video.c
@@ -2553,6 +2553,7 @@ i965_CreateContext(VADriverContextP ctx,
                    VAContextID *context)                /* out */
 {
     struct i965_driver_data *i965 = i965_driver_data(ctx);
+    struct intel_driver_data *intel = intel_driver_data(ctx);
     struct object_config *obj_config = CONFIG(config_id);
     struct object_context *obj_context = NULL;
     VAConfigAttrib *attrib;
@@ -2742,6 +2743,15 @@ i965_CreateContext(VADriverContextP ctx,
 
     i965->current_context_id = contextID;
 
+    i965->bufmgr = drm_intel_bufmgr_gem_init(intel->fd, 4096);
+
+    i965->gem_context = drm_intel_gem_context_create(i965->bufmgr);
+    intel->gem_context = i965->gem_context;
+    intel->gem_ctx_id = i965->gem_context->ctx_id;
+
+    if (kernel_has_gpu_watchdog_support(&i965->intel))
+        i965->intel.has_watchdog = 1;
+
     return vaStatus;
 }
 
@@ -2765,6 +2775,10 @@ i965_DestroyContext(VADriverContextP ctx, VAContextID context)
 
         obj_context->wrapper_context = VA_INVALID_ID;
     }
+
+    drm_intel_gem_context_destroy(i965->gem_context);
+
+    i965->intel.has_watchdog = 0;
 
     i965_destroy_context(&i965->context_heap, (struct object_base *)obj_context);
 
--- a/src/intel_batchbuffer.c
+++ b/src/intel_batchbuffer.c
@@ -30,9 +30,9 @@
 #include <assert.h>
 
 #include "intel_batchbuffer.h"
+#include "i965_drv_video.h"
 
 #define MAX_BATCH_SIZE      0x400000
-
 
 #define LOCAL_I915_EXEC_BSD_MASK        (3<<13)
 #define LOCAL_I915_EXEC_BSD_DEFAULT     (0<<13) /* default ping-pong mode */
@@ -73,7 +73,6 @@ intel_batchbuffer_space(struct intel_batchbuffer *batch)
     return (batch->size - BATCH_RESERVED) - (batch->ptr - batch->map);
 }
 
-
 struct intel_batchbuffer *
 intel_batchbuffer_new(struct intel_driver_data *intel, int flag, int buffer_size)
 {
@@ -98,7 +97,7 @@ intel_batchbuffer_new(struct intel_driver_data *intel, int flag, int buffer_size
     assert(batch);
     batch->intel = intel;
     batch->flag = flag;
-    batch->run = drm_intel_bo_mrb_exec;
+    batch->run = drm_intel_gem_bo_fence_exec;
 
     if (IS_GEN6(intel->device_info) &&
         flag == I915_EXEC_RENDER)
@@ -130,6 +129,7 @@ void
 intel_batchbuffer_flush(struct intel_batchbuffer *batch)
 {
     unsigned int used = batch->ptr - batch->map;
+    int ring_flag = batch->flag & I915_EXEC_RING_MASK;
 
     if (used == 0) {
         return;
@@ -144,7 +144,16 @@ intel_batchbuffer_flush(struct intel_batchbuffer *batch)
     batch->ptr += 4;
     dri_bo_unmap(batch->buffer);
     used = batch->ptr - batch->map;
-    batch->run(batch->buffer, used, 0, 0, 0, batch->flag);
+
+    if (batch->intel->has_watchdog)
+        intel_batchbuffer_configure_watchdog(batch->intel->fd, batch->intel->gem_ctx_id, batch->flag);
+
+    /* FIXME: Add BSD2 ring on supported platforms */
+    if (ring_flag == I915_EXEC_BSD)
+        batch->flag |= LOCAL_I915_EXEC_BSD_RING0;
+
+    batch->run(batch->buffer, batch->intel->gem_context, used, -1, NULL, batch->flag);
+
     intel_batchbuffer_reset(batch, batch->size);
 }
 
--- a/src/intel_batchbuffer.c
+++ b/src/intel_batchbuffer.c
@@ -146,11 +146,17 @@ intel_batchbuffer_flush(struct intel_batchbuffer *batch)
     used = batch->ptr - batch->map;
 
     if (batch->intel->has_watchdog)
-        intel_batchbuffer_configure_watchdog(batch->intel->fd, batch->intel->gem_ctx_id, batch->flag);
+        intel_batchbuffer_configure_watchdog(batch->intel, batch->intel->gem_ctx_id, batch->flag);
 
-    /* FIXME: Add BSD2 ring on supported platforms */
-    if (ring_flag == I915_EXEC_BSD)
-        batch->flag |= LOCAL_I915_EXEC_BSD_RING0;
+    if (batch->intel->has_bsd2) {
+        if (ring_flag == I915_EXEC_BSD)
+            batch->flag |= LOCAL_I915_EXEC_BSD_RING0;
+        else if (ring_flag == (I915_EXEC_BSD_RING2 | I915_EXEC_BSD))
+            batch->flag |= LOCAL_I915_EXEC_BSD_RING1;
+    } else {
+        if (batch->flag == I915_EXEC_BSD)
+            batch->flag |= LOCAL_I915_EXEC_BSD_RING0;
+    }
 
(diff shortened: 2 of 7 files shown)


src/intel_batchbuffer.c
```diff
@@ -98,7 +97,7 @@ intel_batchbuffer_new(struct intel_driver_data *intel, int flag, int buffer_size
     assert(batch);
     batch->intel = intel;
     batch->flag = flag;
-    batch->run = drm_intel_bo_mrb_exec;
+    batch->run = drm_intel_gem_bo_fence_exec;
```
xhaihao: Please update the minimum version requirement for libdrm in configure.ac and menson.build. The current version requirement of 2.4.52 doesn't support drm_intel_gem_bo_fence_exec

src/intel_driver.c
```diff
@@ -56,6 +56,51 @@ uint32_t g_intel_debug_option_flags = 0;
...
+        .param = LOCAL_I915_CONTEXT_PARAM_WATCHDOG,
+        .value = (uint64_t)&threshold,
+        .size = sizeof(threshold)
+    };
+
+    memset(threshold, 0, sizeof(threshold));
+
+    /* set watchdog threshold to VCS's engine only */
+    if (ring_flag == I915_EXEC_BSD || ring_flag == LOCAL_I915_EXEC_VCS2) {
+        threshold[VIDEO_DECODE_CLASS] = 1200;
```
xhaihao: Is 1200 suitable for all media workload?

src/intel_driver.h
```diff
@@ -111,6 +111,18 @@ extern uint32_t g_intel_debug_option_flags;
 #define VA_INTEL_DEBUG_OPTION_BENCH     (1 << 1)
 #define VA_INTEL_DEBUG_OPTION_DUMP_AUB  (1 << 2)
 
+#define LOCAL_I915_CONTEXT_PARAM_WATCHDOG 0x7
+#define LOCAL_I915_EXEC_VEBOX             (4<<0)
```
xhaihao: You may use I915_EXEC_VEBOX which is defined in i915_drm.h

src/intel_driver.h
```diff
@@ -111,6 +111,18 @@ extern uint32_t g_intel_debug_option_flags;
...
+#define LOCAL_I915_EXEC_VCS2              (5<<0)
```
xhaihao: Why is LOCAL_I915_EXEC_VCS2 needed? we don't use 5 (LOCAL_I915_EXEC_VCS2) to indicate vcs2 ring.

src/intel_driver.h
```diff
@@ -183,22 +195,33 @@ struct intel_driver_data {
...
     const struct intel_device_info *device_info;
     unsigned int mocs_state;
 };
 
+/* FIXME: this struct is currently private and part of libdrm */
+struct _drm_intel_context {
+    unsigned int ctx_id;
+    struct _drm_intel_bufmgr *bufmgr;
+};
+
```
xhaihao: Is this structure used in your patch?

src/i965_drv_video.c
```diff
@@ -2742,6 +2743,15 @@ i965_CreateContext(VADriverContextP ctx,
 
     i965->current_context_id = contextID;
 
+    i965->bufmgr = drm_intel_bufmgr_gem_init(intel->fd, 4096);
+
```
xhaihao: drm_intel_bufmgr_gem_init might fail, so a check for i965->bufmgr is needed.

src/i965_drv_video.c
```diff
@@ -2742,6 +2743,15 @@ i965_CreateContext(VADriverContextP ctx,
...
+    i965->gem_context = drm_intel_gem_context_create(i965->bufmgr);
```
xhaihao: drm_intel_gem_context_create might fail too.

src/i965_drv_video.c
```diff
@@ -2766,6 +2776,10 @@ i965_DestroyContext(VADriverContextP ctx, VAContextID context)
         obj_context->wrapper_context = VA_INVALID_ID;
     }
 
+    drm_intel_gem_context_destroy(i965->gem_context);
+
```
xhaihao: drm_intel_bufmgr_destroy() should be called to release i965->bufmgr.

src/intel_batchbuffer.c
```diff
@@ -144,7 +144,16 @@ intel_batchbuffer_flush(struct intel_batchbuffer *batch)
...
     dri_bo_unmap(batch->buffer);
     used = batch->ptr - batch->map;
-    batch->run(batch->buffer, used, 0, 0, 0, batch->flag);
+
+    if (batch->intel->has_watchdog)
+        intel_batchbuffer_configure_watchdog(batch->intel->fd, batch->intel->gem_ctx_id, batch->flag);
+
+    /* FIXME: Add BSD2 ring on supported platforms */
+    if (ring_flag == I915_EXEC_BSD)
+        batch->flag |= LOCAL_I915_EXEC_BSD_RING0;
```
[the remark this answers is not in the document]
cmsanta: Yes, I am planning to add support for BSD2. I will try to get to it.

Yeah, I think when I add support for BSD2 ring that I'll need to add LOCAL_I915_EXEC_BSD_RING1 as part of the batch->flag

src/intel_driver.c
cmsanta: You're right, that's a guestimate for now, I'll try to come up with a better solution that could account for different workloads and different display resolutions, for example, a higher threshold at 4K...

src/i965_drv_video.c
cmsanta: Ok, will take a closer look.

src/i965_drv_video.c
cmsanta: Ok, will take a closer look.

src/intel_driver.h
cmsanta: You're right, I just removed struct _drm_intel_bufmgr and it still compiles just fine. I'll remove it and retest again on my testbed before uploading a new patch version.

src/i965_drv_video.c
cmsanta: Ok, I'll add that and retest.

src/intel_driver.c
tursulin: Would it be possible to do something based on frame rate? However, a) I don't know if that information is available at this level and b) it would probably require distinguishing between batch and real mode which I don't know if it is possible. But for real-time it certainly sounds plausible for a layman like me that if a frame hasn't been decoded in 1/fps time the stack might want to give up and do something else. What happens with inter-frame dependencies is another question I don't have the answer for.

src/intel_driver.c
```diff
@@ -56,6 +56,51 @@ uint32_t g_intel_debug_option_flags = 0;
 #define LOCAL_I915_PARAM_EU_TOTAL 34
 #endif
 
+bool intel_batchbuffer_configure_watchdog(int fd, unsigned int gem_ctx_id, int flag)
+{
+    unsigned threshold[5];
```
tursulin: From the other place this is used I suppose this should be u32. I suppose we will have a look at the uAPI details when the i915 part gets posted.

src/intel_driver.h
```diff
@@ -111,6 +111,18 @@ extern uint32_t g_intel_debug_option_flags;
...
+#define LOCAL_I915_EXEC_VCS2              (5<<0)
+
+typedef enum _gen_gpu_class_engine {
+    RENDER_CLASS = 0,
+    VIDEO_DECODE_CLASS,
+    VIDEO_ENHANCEMENT_CLASS,
+    COPY_ENGINE_CLASS,
+    GEN_GPU_ENGINE_NUMBER
+} _gen_gpu_class_engine, *pgen_gpu_class_engine;
+
```
tursulin: Above looks a copy of enum drm_i915_gem_engine_class from i915_drm.h, apart from the last entry which I don't understand.

src/intel_driver.h
tursulin: Also, i915 does not and will not support this flag. If you were thinking of adding support for new engines you will have to make use of the context engine map feature (or the Virtual Engine). Neither is upstream yet.

src/intel_driver.h
cmsanta: I took it from an internal patch from a different implementation for the iHD media driver thinking that was just a simple list of enums of the different type of engines however, using the struct from i915_drm.h after bumping libdrm to version 4.92 that includes this:

/*
 * Different engines serve different roles, and there may be more than one
 * engine serving each role. enum drm_i915_gem_engine_class provides a
 * classification of the role of the engine, which may be used when requesting
 * operations to be performed on a certain subset of engines, or for providing
 * information about that group.
 */
enum drm_i915_gem_engine_class {
	I915_ENGINE_CLASS_RENDER	= 0,
	I915_ENGINE_CLASS_COPY		= 1,
	I915_ENGINE_CLASS_VIDEO		= 2,
	I915_ENGINE_CLASS_VIDEO_ENHANCE	= 3,

	I915_ENGINE_CLASS_INVALID	= -1
};

I end up with a GPU hang. Note that the struct are similars but not identical, specifically, VIDEO_DECODE_CLASS is 1 on my definition whereas I915_ENGINE_CLASS_VIDEO is 2 in i915_drm.h.

src/intel_driver.h
cmsanta: @tursulin which flag are u referring to? the one being suggested by @xhaihao  (I915_EXEC_BSD_RING2 | I915_EXEC_BSD? I tried on my testbed and it both compiles and executes correctly on my testbed.

src/i965_drv_video.c
cmsanta: @xhaihao please take a look at updated patch on this branch.

src/i965_drv_video.c
cmsanta: @xhaihao please take a look at updated patch on this branch.

src/intel_driver.h
cmsanta: @xhaihao please take a look at updated patch on this branch.

src/i965_drv_video.c
```diff
@@ -2744,14 +2744,18 @@ i965_CreateContext(VADriverContextP ctx,
     i965->current_context_id = contextID;
 
     i965->bufmgr = drm_intel_bufmgr_gem_init(intel->fd, 4096);
-
+    assert(i965->bufmgr);
```
xhaihao: Please return error instead of assert if failing to initialize the buffer management

src/i965_drv_video.c
```diff
@@ -2744,14 +2744,18 @@ i965_CreateContext(VADriverContextP ctx,
...
     i965->gem_context = drm_intel_gem_context_create(i965->bufmgr);
+    assert(i965->gem_context);
```
xhaihao: Please return error instead of assert if failing to create hw the context

src/intel_batchbuffer.c
```diff
@@ -146,11 +146,17 @@ intel_batchbuffer_flush(struct intel_batchbuffer *batch)
...
     if (batch->intel->has_watchdog)
-        intel_batchbuffer_configure_watchdog(batch->intel->fd, batch->intel->gem_ctx_id, batch->flag);
+        intel_batchbuffer_configure_watchdog(batch->intel, batch->intel->gem_ctx_id, batch->flag);
 
-    /* FIXME: Add BSD2 ring on supported platforms */
-    if (ring_flag == I915_EXEC_BSD)
-        batch->flag |= LOCAL_I915_EXEC_BSD_RING0;
+    if (batch->intel->has_bsd2) {
+        if (ring_flag == I915_EXEC_BSD)
+            batch->flag |= LOCAL_I915_EXEC_BSD_RING0;
```
xhaihao: The KMD driver will use the ping-pong mode to select the right ring if neitherLOCAL_I915_EXEC_BSD_RING0 nor LOCAL_I915_EXEC_BSD_RING1 is set.  Specifying LOCAL_I915_EXEC_BSD_RING0 will cause performance drop for multiple decoding/encoding cases.

BTW LOCAL_I915_EXEC_BSD_RING0 is equal to I915_EXEC_BSD_RING1 and LOCAL_I915_EXEC_BSD_RING1 is equal to I915_EXEC_BSD_RING2, but I915_EXEC_BSD_RING1 and I915_EXEC_BSD_RING2 were not defined in older version of libdrm, so we added LOCAL_I915_EXEC_BSD_RING0 and LOCAL_I915_EXEC_BSD_RING1 in the driver.

src/intel_batchbuffer.c
```diff
@@ -146,11 +146,17 @@ intel_batchbuffer_flush(struct intel_batchbuffer *batch)
...
+        else if (ring_flag == (I915_EXEC_BSD_RING2 | I915_EXEC_BSD))
```
xhaihao: ring_flag = batch->flag & I915_EXEC_RING_MASK, so it is always false for `ring_flag == (I915_EXEC_BSD_RING2 | I915_EXEC_BSD`

src/intel_driver.c
```diff
@@ -73,11 +73,11 @@ bool intel_batchbuffer_configure_watchdog(int fd, unsigned int gem_ctx_id, int f
     memset(threshold, 0, sizeof(threshold));
 
     /* set watchdog threshold to VCS's engine only */
-    if (ring_flag == I915_EXEC_BSD || ring_flag == LOCAL_I915_EXEC_VCS2) {
-        threshold[VIDEO_DECODE_CLASS] = 1200;
+    if (ring_flag == I915_EXEC_BSD || ring_flag == (I915_EXEC_BSD_RING2 | I915_EXEC_BSD)) {
```
xhaihao: ring_flag = flag & I915_EXEC_RING_MASK, so it is always false for ring_flag == (I915_EXEC_BSD_RING2 | I915_EXEC_BSD

kwizart:
I wish this PR could be finished as a watchdog feature for this legacy driver would be very welcomed.

uartie:
No activity.  Closing.

@cmsanta please reopen if you plan to continue to work on this.

Otherwise, of someone else wants to finish this, feel free to open a new pull-request with your updates.

PR #441, opened 2019-01-27
Fix meson build without wayland

Erick555:
Current 2.3.0 build with meson fails if wayland dependency isn't detected:
```
        Library dl found: YES
        Library m found: YES
        Program git found: NO
        Dependency threads found: YES 
        Found pkg-config: /usr/bin/pkg-config (1.5.3)
        Dependency libdrm found: YES 2.4.97
        Dependency libdrm_intel found: YES 2.4.97
        Dependency libva found: YES 1.3.0
        Dependency libva-x11 found: YES 1.3.0
        Found CMake: NO
        Dependency wayland-client found: NO (tried pkgconfig and cmake)
        Dependency libva-wayland found: NO (tried pkgconfig and cmake)

        meson.build:94:2: ERROR:  Unknown variable "wl_scanner".
```

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=235039

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -91,7 +91,7 @@ if get_option('with_wayland') != 'no'
     version : libva_version,
     required : get_option('with_wayland') == 'yes')
 
-  WITH_WAYLAND = wl_scanner.found() and libva_wayland_dep.found()
+  WITH_WAYLAND = wayland_client_dep.found() and libva_wayland_dep.found()
 endif
 
 subdir('src')


meson.build
```diff
@@ -91,7 +91,7 @@ if get_option('with_wayland') != 'no'
     version : libva_version,
     required : get_option('with_wayland') == 'yes')
 
-  WITH_WAYLAND = wl_scanner.found() and libva_wayland_dep.found()
```
ceyusa: Perhaps, a better fix is to move out the wl_scanner's find_program() out of the conditional, and add the wayland_client_dep.found() in the WITH_WAYLAND variable, keeping the wl_scanner.found()

Erick555: @ceyusa I'm not much familiar with these mechanics. Would you mind to open new PR with correct fix? I'll close this one then.

Erick555: @ceyusa Thx.

PR #471, opened 2019-06-03
Postpone the update of vp9_frame_ctx

xhaihao:
Mapping the buffer for vp9_frame_ctx at once after batch buffer
submission will cause GT frequency boost, so postpone the update of
vp9_frame_ctx to next frame

Signed-off-by: Haihao Xiang <haihao.xiang@intel.com>
Tested-by: Shankar, Vaibhav <vaibhav.shankar@intel.com>


lizhong1008:
LGTM

PR #474, opened 2019-06-14
Green artefacts in WVC1 decoding with sandy bridge (gen6)

J-Bu:
Since commit 1591f6c1be689c5265cb5beab18c97771d056cb9 WVC1 videos are decoded with a green color on sandy bridge platforms.
This is because of a wrong ptype before this commit ptype was set to GEN6_VC1_BI_PICTURE if picture_type is GEN6_VC1_I_PICTURE and profile is advanced. With this commit the ptype is correctly set to GEN6_VC1_BI_PICTURE for advanced profiles.

Maybe this is also needed for the other platforms affected by commit 1591f6c1be689c5265cb5beab18c97771d056cb9  but I have no hardware to test it.

### DIFF
--- a/src/gen6_mfd.c
+++ b/src/gen6_mfd.c
@@ -1482,8 +1482,11 @@ gen6_mfd_vc1_pic_state(VADriverContextP ctx,
         loopfilter = pic_param->entrypoint_fields.bits.loopfilter;
     }
 
-    if (picture_type == GEN6_VC1_I_PICTURE || picture_type == GEN6_VC1_BI_PICTURE) /* I picture */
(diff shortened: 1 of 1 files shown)


xhaihao:
Thanks for the patch. If possible, could you share your test clip?

J-Bu:
I uploaded two videos where the error occurs please use following link to download:
https://cloud.unicon-software.com/d/e5eb570bf70843049200/

xhaihao:
I don't see green color with commit https://github.com/intel/intel-vaapi-driver/commit/1591f6c1be689c5265cb5beab18c97771d056cb9 on other platforms.

PR #491, opened 2019-10-09
Check for python3 rather than python2

bryteise:
With python2 being deprecated, it makes sense to switch gpp.py to be
python3 compatable and the build system to look for python3 rather
than python2.


configure.ac
```diff
@@ -77,7 +77,7 @@ AC_HEADER_STDC
 AC_SYS_LARGEFILE
 AC_CHECK_LIB([m], [sin])
 AC_CHECK_FUNCS([log2f])
-AC_CHECK_PROGS([PYTHON2], [python2 python])
+AC_CHECK_PROGS([PYTHON3], [python3 python])
```
xhaihao: PYTHON2 is used in some Makefile.am files, does it not need to be replaced with PYTHON3?

bryteise: Ah I didn't notice those, thanks! I have updated the PR. For my use case it doesn't seem like those shaders were being built so I didn't see any error in my testing, sorry about that.

PR #497, opened 2019-12-09
build: Allow clean build on OpenBSD 6.6.

iarebatman:
We don't need to link with -ldl on OpenBSD.
Add new check to see if we need to find a libdl
before we assume we do.

swap32 is provided by OpenBSD.  Use it when
available.

### DIFF
--- a/src/i965_encoder_utils.c
+++ b/src/i965_encoder_utils.c
@@ -82,6 +82,7 @@ struct __avc_bitstream {
 
 typedef struct __avc_bitstream avc_bitstream;
 
+#ifndef __BSD_VISIBLE
 static unsigned int
 swap32(unsigned int val)
 {
@@ -92,6 +93,7 @@ swap32(unsigned int val)
             (pval[2] << 8)      |
             (pval[3] << 0));
 }
+#endif
 
 static void
 avc_bitstream_start(avc_bitstream *bs)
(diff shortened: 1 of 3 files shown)


src/i965_encoder_utils.c
```diff
@@ -82,6 +82,7 @@ struct __avc_bitstream {
 
 typedef struct __avc_bitstream avc_bitstream;
 
+#ifndef __BSD_VISIBLE
```
iamleot: I think that it's safer to change that to:

````
#ifndef __OpenBSD__
````

...or properly check for `swap32()` existence in an OS-agnostic way.

(From a quick skim both DragonFlyBSD and FreeBSD defines `__BSD_VISIBLE` but none of them has a `swap32(3)` function and if that's applied I think that would break `intel-vaapi-driver` there.)

jbeich:
Can you apply `-ldl` change to autotools build as well? `AC_SEARCH_LIBS([dlopen], [dl])` automatically appends `-ldl` to `LIBS`, so you can just remove explicit `-ldl` from *.am files.

Also, `--enable-hybrid-codec` uses `RTLD_NODELETE` which is missing on OpenBSD.

uartie:
No response to reviews from @iarebatman.  Closing

Please reopen and update if you plan to address the reviewers.

PR #511, opened 2020-05-22
Add FreeBSD to CI

jbeich:
Depends on https://github.com/intel/libva/pull/363


jbeich:
Oops, forgot to mark as a draft. 597908978d73 is not supposed to be merged, only added for others to inspect **green** TravisCI build.

PR #514, opened 2020-07-05
fix exporting buffers with 3 planes and VA_EXPORT_SURFACE_SEPARATE_LAYERS

michaelolbrich:
To get the plane offset, y_cb_offset or y_cr_offset must be multiplied with
the pitch for the 'Y' plane. See for example i965_DeriveImage().

Without this, exporting formats with 3 planes where cb_cr_pitch == width/2
is broken, because the offset of the third plane is calculated incorrectly.

### DIFF
--- a/src/i965_drv_video.c
+++ b/src/i965_drv_video.c
@@ -6904,16 +6904,17 @@ i965_ExportSurfaceHandle(VADriverContextP ctx, VASurfaceID surface_id,
                 else
                   y_offset = obj_surface->y_cr_offset;
             } else {
-                y_offset = obj_surface->y_cr_offset - obj_surface->y_cb_offset;              
-                if (y_offset < 0)
-                  y_offset = -y_offset;
+                if (obj_surface->y_cb_offset < obj_surface->y_cr_offset)
+                  y_offset = obj_surface->y_cr_offset;
+                else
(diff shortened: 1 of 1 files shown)


rmader:
Just wanted to drop that I've been hitting this as well. I've worked around it with [a dirty Gstreamer patch](https://gitlab.freedesktop.org/rmader/gstreamer/-/commit/776628b19875b7b21a63ca9fd9673c295c86e97a) which usually gives the right results, but not always. Would be nice to still see this land!

rmader:
P.S.: I verified that this MR fixes playback with Gstreamer 1.24 `vampeg2dec` on my Thinkpad T400 (GM45/P8400) with the following samples:
 - https://kodi.wiki/view/Samples : [MPEG2 1080i/29.97](https://drive.google.com/file/d/0BwxFVkl63-lETXlOa3hEMmVyZGM/view?usp=sharing&resourcekey=0-FG5JbjCsd8QduRhkZ2aT8A) (U.S. ATSC OTA capture w/DD5.1 audio, muxed into MKV)
 - https://docs.evostream.com/2.0/userguide_samplecontent.html [bun33s.ts](http://docs.evostream.com/sample_content/assets/bun33s.ts)

PR #530, opened 2021-02-09
Fix JPEG decoding of an oddly encoded files on gen7/8

digetx:
Intel VA-API driver fails to decode couple JPEG files for me because of the offending implicit u8->int32 expansion in the code of the driver.  This PR fixes the implicit integer expansion and decoding works properly now.


rathann:
@digetx could you provide a sample JPEG file that fails to decode? I have a Haswell CPU and cannot find any JPEGs that would fail to decode locally.

PR #542, opened 2021-11-23
Update README

martin-kokos:
Add notice of intel media driver

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,8 @@
 
 VA-API (Video Acceleration API) user mode driver for Intel GEN Graphics family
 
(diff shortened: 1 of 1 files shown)


martin-kokos:
> Maintenance mode is probably more correct? Like.. it's not that they stopped development per se (like perhaps intel-hybrid-driver). It's more that there's nothing left to add/expand.
> 
> p.s. intel's own mailing lists are no more, at most the readme should point to the [archive thereof](https://web.archive.org/web/20220916014841/https://lists.01.org/hyperkitty/list/intel-vaapi-media@lists.01.org/)

That's what I meant. People looking for vaapi support on intel shouldn't be installing it and rather look at https://github.com/intel/media-driver instead id they can.

mirh:
Right, I forgot the part where you do actually have a choice to push ahead since non-Atom Gen8 hardware. For as much as I don't think <Gen12 is seeing much love there either (except perhaps in the common bits for HDR or whatever other fancier new modernity)

PR #549, opened 2022-05-02
Meson warning fixups

eli-schwartz:


eli-schwartz:
Setting the `check: ` kwarg requires newer Meson. I if/elsed the code for now as I'm not sure what the project policy is. I guess bumping it would drop support for Ubuntu 18.04 which probably should not matter.

ceyusa:
libva already uses `meson_version : '>= 0.53.0'` : https://github.com/intel/libva/blob/master/meson.build#L11 

Perhaps it would worth to bump to the same version or, at least to 0.47

eli-schwartz:
Good point, done.

eli-schwartz:
I would be happy to receive clarification on that as well.

